PR #139, opened 2014-08-15
Public events view improvements

russss:
Improve the public events view:

* Include the event type in speakers.json
* Display un-scheduled (but confirmed) events in the "events" view - they were already displayed in the "speakers" view
* Render event pages for un-scheduled (but confirmed) events
* Add a JSON endpoint for the "events" view


manno:
I'm closing this PR, since most features have been merged already and smaller pull requests for single issues are way easier to handle.

Regarding the remaining point, unscheduled events, we can discuss that in issues #177 and #191

PR #150, opened 2014-12-30
let the organiser use markdown for this text

automatical:
This change will allow conference administrators to use Markdown if they wish for the main CFP text.

HTML is still supported as the markdown parser ignores this.

### DIFF
--- a/test/unit/notification_test.rb
+++ b/test/unit/notification_test.rb
@@ -4,8 +4,8 @@ class NotificationTest < ActiveSupport::TestCase
 
   setup do
     @cfp = FactoryGirl.create(:call_for_papers)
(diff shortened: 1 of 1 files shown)


app/views/cfp/sessions/new.html.haml
```diff
@@ -14,7 +14,7 @@
       = render "shared/flash", :flash => flash
       - if @conference.call_for_papers.welcome_text
         %p
-          = simple_format @conference.call_for_papers.welcome_text
+          = simple_format render_markdown(@conference.call_for_papers.welcome_text)
```
manno: I'm not sure about using simple_format together with render_markdown?

PR #151, opened 2014-12-30
enable datepickers on the new CFP page

automatical:
jQuery datepickers aren't in use here, I thought it made sense to add them.


manno:
There have been some changes to datepickers, because of simple_form, but yes would make total sense.

manno:
This has been implemented in #232

PR #153, opened 2015-01-12
Shorten slugs to usable length, fix #131

### LINKED ISSUE #31: edit conference: max timeslots off by 1?
Hi,

I have a timeslot duration of 00:05.
When I set default timeslot to 3 and max timeslots to 3, and go to 'create event', the default is 0:10 and the only options are '0:00, 0:05, 0:10'.

Increasing max to 4 gives as options '0:00, 0:05, 0:10, 0:15'

Probably, the option of a talk of 00:00 time (which is a surprising option) is also counted?

### LINKED ISSUE #131: fix slug generation
In my opinion ```gpn14_-_5853_-_de_-_medientheater_-_201406221300_-_spaceautomatisierung_-_schinken``` ist not a really good slug. We should discuss alternatives and implement the winner.

My preference is ```gpn14_5853_de-schinken-spaceautomatisierung```

Still unsolved issues: 
* what if a file contains multiple audio tracks, e.g. a translated one. We used ```gpn14-5853-de-en-spaceautomatisierung``` in the past for this cases. Currently only the C3-Tracker (FeM/VOC) knows which talk was translated and which not
* multiple speakers e.g. five? -> drop the speaker name from the slug generation, if there are more than two speakers?

jjeising:
This changes the public slug format from `31c3_-_6561_-_en_-_saal_1_-_201412271100_-_31c3_opening_event_-_erdgeist
_-_geraldine_de_bastion` to `31c3-6561-en-31C3_Opening_Event ` a format similar to the filenames published in the recent years, but without a language identifier.

The original language is already available in the schedule and can be appended if needed. Published recordings may contain other languages as translations so the original
language should be excluded from the slug.

It uses truncate from TextHelper to enforce a maximum length. This shouldn’t be an issue in general as event.title has the default string column limit of 255 chars, but strange conference acronyms may break that.

Calls to_str to get an unsafe string for HTML templates.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,5 +1,6 @@
 class Event < ActiveRecord::Base
   include ActiveRecord::Transitions
+  include ActionView::Helpers::TextHelper
 
   before_create :generate_guid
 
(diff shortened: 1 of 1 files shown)


saerdnaer:
@manno speaks anything against an merge?

manno:
Has been merged in 12437c6

PR #160, opened 2015-03-20
Allow heroku deployments

bennibu:
To make a deployment via heroku or other PAAS provider easier, I have

* replaced the config/settings.yml behaviour with ENV variables. See `env.example`
* added the rails_12factor gem to serve assets in production and log to STDOUT.
* add puma as default production server with one worker and 5 threads as default
* add s3 support for paperlip uploads to amazon s3 (not tested yet)
* set ruby version to 2.1.5 (see `Gemfile`)

I appreciate any comments.


brianredbeard:
@bennibu to make the tests pass you would need to remove the ruby 1.9.3 handling done in `travis.yml`.

bennibu:
hey brian, I removed the ruby definition in Gemfile to get travis ci working. The last [builds](https://travis-ci.org/frab/frab/builds/57933007) succeeded..

zonque:
Nice feature addition! A couple of remarks:

* What's the motivation for the change from `WEBrick` to `puma`? After all, the docs already clearly state that for production, you should consider something like `unicorn`? In any case, this commit seems unrelated to the goal of this patch set.
* Consider adding an application manifest to describe the parameters needed for the setup. This also allows you to automatically generate keys through `"generator": "secret"`. See the [Heroku Docs](https://blog.heroku.com/archives/2014/5/22/introducing_the_app_json_application_manifest)
* AFAIK, the ruby version detail is merely a suggestion made by Heroku, rather than a hard requirement
* In order to avoid name conflicts, it might be a good idea to prefix /all/ environment variables with `FRAB_`. After all, you cannot easily do that later, as deployed versions would need to update as well.
* I guess there should be a way to migrate existing installations by reading `settings.yml` and writing `.env`

manno:
With Rails 4 moved to dotenv and puma, which should make heroku deployments easier. 

I'm closing this PR since most frab installations I know are self-hosted. I don't want to break existing deployments by adding s3 and 12factor. If there are any changes in here that would make sense for both deployment scenarios, please feel free to create separate pull requests.

PR #164, opened 2015-05-05
Techrider

zonque:
Techrider is a free-form field similar to `Notes` that's supposed to be used to specifiy resources that are needed for an event. Examples are: projector, PA, microphone, a bottle of champagne ...

This PR contains a migration, updated views and a special-purpose report for all events with a Techrider.

### DIFF
--- a/app/controllers/cfp/events_controller.rb
+++ b/app/controllers/cfp/events_controller.rb
@@ -106,7 +106,7 @@ class Cfp::EventsController < ApplicationController
 
   def event_params
     params.require(:event).permit(
-      :title, :subtitle, :event_type, :time_slots, :language, :abstract, :description, :logo, :track_id, :submission_note,
+      :title, :subtitle, :event_type, :time_slots, :language, :abstract, :description, :logo, :track_id, :submission_note, :tech_rider,
       event_attachments_attributes: %i(id title attachment public _destroy),
       links_attributes: %i(id title url _destroy)
     )
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -222,7 +222,7 @@ class EventsController < ApplicationController
 
   def event_params
     params.require(:event).permit(
-      :id, :title, :subtitle, :event_type, :time_slots, :state, :start_time, :public, :language, :abstract, :description, :logo, :track_id, :room_id, :note, :submission_note, :do_not_record, :recording_license,
+      :id, :title, :subtitle, :event_type, :time_slots, :state, :start_time, :public, :language, :abstract, :description, :logo, :track_id, :room_id, :note, :submission_note, :do_not_record, :recording_license, :tech_rider,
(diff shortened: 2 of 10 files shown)


zonque:
Just refreshed this one with some minor nits.

manno:
I would rather not add another field to the submission form. Can't this go into 'submission notes'? 

@ScottyTM But sure, if enough conferences need this?

ScottyTM:
I think this is what the submission notes field is for.

meise:
I think a dedicated field or something else would be very helpful, to get a better overview of speakers requirements. Look for example into the @froscon 2015 submissions. 81 of 201 have a more or less useful submission note. Reviewing all the notes costs a lot of time, not only for the program team.

Another solution for this problem could be to add check boxes, that a speaker can specify what kind of information was added to the notes field.

manno:
Honestly, I think with a second field you just got two problems instead of one. Now you have to read both of them. You just can't trust submitters to fill out the form correctly and be done with it. You have to ask them and you have to check their inputs. FrOSCon did this very well. 

I understand the need to have this data in a more accessible form, before the conference starts and while it's running. But I don't think that information can be collected from the submitters. Maybe have the content crew collect this information from emails and the submission notes field and put it in the admin notes?

zonque:
> I understand the need to have this data in a more accessible form, before the conference starts and while it's running. But I don't think that information can be collected from the submitters. Maybe have the content crew collect this information from emails and the submission notes field and put it in the admin notes?

The admin notes don't suffice either. On our conference, we needed a list of things that are needed to run the events upcoming in the next two hours, so you can send people taking care of it. Handing out all other notes would require those people to parse larger amounts of texts.

One solution about not trusting speakers to carefully fill out that information would be to only make that field accessible by the organizers, so it can be maintained with care.

manno:
I was under the impression no one uses the admin_notes field for anything useful... But ok, if one admin notes field is not enough, let's add another one for resources.

saerdnaer:
@manno Can you merge this PR, so we can use it at 32C3?

zonque:
@saerdnaer this cannot be merged automatically right now. If you need that for 32c3, I can have a look and rebase this change. I actually wanted to do this anyway, so this might be a good incentive :) Any other feature in pending PR you're interested in?

saerdnaer:
@zonque I didn't see any pull request, but there are some other issues/ideas. Just ask when you are one with this one... ;-) 
For now: It would be nice if the event report can also list the actual content or the tech rider field, so the user don't have to click through all listed events.

zonque:
Ok, rebased now, please have a look!

manno:
I merged this too, but it removes the 'do_not_record' checkbox. If this is this intentional I can revert the restore-commit cb9641d?

https://github.com/frab/frab/pull/164/files#diff-61eb47a6ac75629cf64b301842873a22L32

zonque:
Oops, no, that wasn't intentional. Thanks for spotting and fixing up! :)

PR #165, opened 2015-05-05
Expenses

zonque:
Add a data model to store expenses for people of a conference.

This PR adds a new model, its views, and a report to show which non-reimbursed expenses for a conference exist.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,3 +1,5 @@
 FRAB_HOST=localhost
 FRAB_PROTOCOL=http
 FROM_EMAIL=frab@localhost
+#FRAB_CURRENCY_UNIT="€"
+#FRAB_CURRENCY_FORMAT="%n%u"
--- a/app/controllers/expenses_controller.rb
+++ b/app/controllers/expenses_controller.rb
@@ -1,0 +1,50 @@
+class ExpensesController < ApplicationController
+
+  before_filter :authenticate_user!
+  before_filter :not_submitter!
+  before_filter :find_person
+
+  def new
+    @expense = Expense.new
+    flash[:alert] = "#{@person.full_name} does not currently have any expenses."
+  end
+
+  def edit
+    @expense = @person.expenses.find(params[:id])
+  end
+
+  def index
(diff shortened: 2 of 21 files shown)


zonque:
Pushed a rebased version that also contains a simple unit test.

manno:
The conferences I know explicitly decided against such a field in the past, since reimbursements are not very common.

sebix:
We also don't do it because then many people would request a reimbursement without needing it and straining the budget. Those who really need it, ask via mail or direct contact.

If this feature can be deactivated without changes in the code, I'm not against it (can't interpret the code if this is the case).

zonque:
> The conferences I know explicitly decided against such a field in the past, since reimbursements are not very common.

Well, and we decided for it, for valid reasons. For our conference, we need a way to keep track about expenses we eventually need to reimburse to speakers, hence I added this feature. If you're concerned about people adding unjustified expenses themselves, this can be made a conference organizer feature only (not sure how I implemented it back then).

In general, all the PRs I opened add features that are crucial for us in order to run our conference. Things that don't find their way upstream will need rebasing on our side, which is extra work I'd like to avoid.

manno:
As far as I can see this does not add any fields to the CfP views. Introducing this for conference organizers only sound good.

zonque:
Ok, will have a look at what's left of these PRs once the dust settles and I find some time to continue working on all that. It might take weeks or months though.

zonque:
I just rebased this branch against current master and refactored all necessary bits to Rails4, dotenv, SimpleForm etc.

As mentioned earlier, this does not contain a CfP form addition, so it's considered a feature for administrators only.

Works fine in my tests. Please have a look!

manno:
Thanks, I merged the feature. I put the schema.rb changes in a separate commit to avoid merge conflicts. I'm not sure if reviewers should be able to view expenses, but we can deal with that later.

zonque:
Yes, the role model is questionable. It should probably be dealt with in a separate PR or so.

PR #168, opened 2015-05-06
event: show subtitle in _table partial

zonque:
Display the event subtitle in the event overview.


zonque:
This is now included in #182. Closing.

Issue #107, opened 2013-10-07
Search Index Corruption


manno:
The search index of `act_as_indexed` becomes corrupt, resulting in missing/wrong search results.

This happens after time, or even after 'big' changes to a conference.

Solution: remove the index directory `tmp/index`.

PR #171, opened 2015-05-07
RFC: Use Ransack for searches through web forms

zonque:
The results provided for web search requests require very strict matches as they go through acts_as_indexed. Ransack is already in the list of gems but only used when manual external requests are used with q= parameter assignment, using the Ransack query format.

Also, according to #107, acts_as_index causes index corruption.

To fix this, this PR implements a better way of searching by using Ransack with 'cond' predicates on the fields of events, people and conferences. In a 2nd step, acts_as_indexed is removed all together.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gem 'pg', group: :postgresql
 gem 'sqlite3', group: :sqlite3
 
 gem 'acts_as_commentable'
-gem 'acts_as_indexed'
 gem 'barista'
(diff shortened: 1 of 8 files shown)


manno:
Hi, thanks for your contributions. I'll try to find some time to merge the pull requests after the Chaos Cologne (17 May) is over.

zonque:
No problem, I'm in no hurry for all this. Our congress is end of May, and I'll probably push more PRs until that so the changes don't get lost :)

manno:
Merged

PR #172, opened 2015-05-09
Update languages.pt_BR.yml

Purodha:
Add Colognian language with its ISO 639 code ksh


manno:
There is no actual translation into Colognian in this pull request.

PR #180, opened 2015-07-08
Event: some assorted improvements

zonque:
Some patches that I hacked on during a conference, and which improved our workflow a bit.


manno:
I like the start_time clean up. However I'm unsure about more event types. But, event types are mostly used in views, only the report controller seems to use them in a query. So why not.

manno:
I merged 30dabcd and 9f2d7c2, feel free to open a new PR for the start times.

PR #181, opened 2015-07-08
Add mass mail feature

zonque:
This pull request implements a simple mass mail feature that we used during a conference to contact confirmed speakers with latest updates. The filter mechanism that controls the list of receivers is easily extendable. It also includes a simple template engine that allows a bit of mail content personification.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,7 @@ group :development, :test do
   gem 'bullet'
   gem 'pry-rails'
   gem 'pry-byebug'
+  gem 'letter_opener'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,6 +45,7 @@ GEM
       thread_safe (~> 0.3, >= 0.3.4)
       tzinfo (~> 1.1)
     acts_as_commentable (4.0.2)
+    addressable (2.4.0)
     arel (6.0.3)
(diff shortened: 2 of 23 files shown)


sebix:
As I don't see anything referring to SMTP, how are mails sent? Directly or via the local mailer, e.g. `mail` or `sendmail`?

zonque:
That is up to the Rails environments settings and is configured once for all occasions in which mail is sent.

This PR also adds `letter_opener`, a gem that opens mail in the default web browser instead of sending it to a mail server, when the server is running in `development` environment.

sebix:
Would be nice if the configuration would be documented somewhere.

zonque:
Rebased and slightly refactored again. Also added a commit for unit testing.

manno:
This is great. Is database cleaner really a speed improvement for our test suite?

zonque:
So, what's the way forward in this case? Why did you cherry-pick only one out of three commits, and how would you proceed with the other two?

manno:
The commit was easy to pick and I wanted to add database_cleaner. The other two need rebasing, that's why I added the 'scheduled' tag. It's kind of scheduled for merging.

zonque:
I rebased this branch against origin/master now. Seems to work just fine. Feedback appreciated :)

manno:
Merged too

PR #182, opened 2015-07-09
Assorted UI improvements

zonque:
Here are some UI related patches that were collected during a conference.


ScottyTM:
Do you have some screenshots for this?

ScottyTM:
Do you have some screenshots for this?

zonque:
No, I don't. Just check out the branch and have a look.

app/views/public/schedule/style.css.erb
```diff
@@ -103,5 +103,5 @@
 /* Custom Conference CSS */
 
 <% if @conference.schedule_custom_css.present? %>
-  <%= @conference.schedule_custom_css.gsub(/\{([^}]*)\}/m){ '{' + sanitize_css($1) + '}'; } %>
```
manno: While this should go into a helper, we do need to sanitize user supplied css.

zonque: Why? This field is only writable by admins, which have numerous other ways of supplying malicious content. Also, as I wrote in my commit logs, the naive cleanup method used in this case messes with more complex CSS.

manno: There should be no way for frab admins to introduce malicious content.

zonque: As I said, the code as it stands tampers with perfectly valid, non-malicious content. So if frab wants to stick with this kind of sanitizing, it needs to be improved a lot and replaced by something that actually understands the semantics of CSS.

I'll keep that commit in my tree as that's currently the easiest way to make frab to what I want it to.

PR #183, opened 2015-07-09
events: add action to copy an event from one conference to another

zonque:
When an event can't happen in one year due to time limitiations, and it is
already created, it is very handy to have a way to simply make a duplicate
of the event and store it into another conference.

This is what this patch implements. The associated people and the
attachments are also duplicated, but the rest is left unset. The exising
(old) record is not altered in any way, the new record is a full clone of
the old one.


sebix:
As I can't read/interpret the Ruby code: Is the button in the event view available for both the submitter and admins (preferring all)? Is it available for any state or only for withdrawn (I prefer it for all states)?

You say, the people are duplicated. Do you mean the peoples account are duplicated? (preferring not).

zonque:
Everyone who can see events in the origin conference and who can create events in the target conference may use this feature. I didn't add any filter for the state of the origin event, so all should be possible.

The people are not duplicated but only their relation to events. That's called `EventPeople` in the data model.

This is just a helper shortcut for what you would normally do with a dozen clicks. I added it because this operation was a common case in a conference I helper organize.

sebix:
Okay, great. Thanks for all your efforts btw!

This PR implements, as also the most others, a very handy feature which I missed already some times.

zonque:
Welcome. I have some more things in the queue, but they need rebasing. So I'll wait until some of the currently open PRs are merged.

app/controllers/events_controller.rb
```diff
@@ -116,6 +116,7 @@ def new
   def edit
     @event = Event.find(params[:id])
     authorize! :update, @event
+    @other_conferences = Conference.where("id != ?", @event.conference.id)
```
manno: Maybe limit to most recent 10 conferences...

app/controllers/events_controller.rb
```diff
@@ -200,6 +201,24 @@ def destroy
...
 
+  def copy_to_conference
+    old_event = Event.find(params[:id])
+    new_event = old_event.dup
+    authorize! :create, new_event
+
+    new_conference = Conference.find(params[:event][:conference_id])
+    new_event.conference = new_conference
+    new_event.event_attachments << old_event.event_attachments
+    new_event.event_people << old_event.event_people
```
manno: This actually creates new event_people?

manno:
I'm unsure if this is really needed?

app/controllers/events_controller.rb
sebix: (ruby noob here) Does this also give conferences with no current CfP?

sebix:
At least I'm for it. We (organizers, but also submitters) had several use cases for such a feature for our last conference.

app/controllers/events_controller.rb
ofeldt: `ActiveRecord::Base#dup` creates a shallow copy, clearing associations. I'd say it will just link all `old_event.event_*` to the new event. Thus it is **not a deep clone** and would mess with conferences history/state by *"stealing"* associated objects which hold a foreign key to a conference.

zonque:
> I'm unsure if this feature is really needed?

On the conference I'm helping with, we constantly had the problem of speakers eventually being unavailable, after we already added their database entries. Then, the year after, we'd like to consider them again and put them on the list, by copying data over from an older conference.

So, yes, for us, this feature is really needed. But I can as well keep it local to my fork of course.

zonque:
Not sure about this one - is it considered useful? I'm happy to rebase it then. Otherwise, I'll keep it as out-of-tree patch for my installations, as my users requested it :)

zonque:
Now rebased to current master. Please consider merging.

app/views/events/edit.html.haml
```diff
@@ -9,3 +9,11 @@
   .row
     .span16
       = render 'form'
+  - if @other_conferences.any?
+    = semantic_form_for(@event, :url => copy_to_conference_event_path) do |f|
```
zonque: This doesn't work anymore since the rebase. Don't merge this yet, will push a new version.

zonque:
This is rebased now, along with the change that associations are also fully duplicated, so the original event should remain intact. Thanks to @ofeldt for the heads-up.

manno:
Ok, people ask for this all the time, so maybe it's needed after all. However I'd feel more comfortable merging this if there was a test for the `copy_to_conference` action.

Since `dup` does not `deep_dub`, this would create a new event in another conference, with the room still set to the original conferences rooms. Event also belongs to `track`, so same problem there.

zonque:
Ah, right, I was thinking about unscheduled events only.

I will add a test for this, and switch to deep duplication.

manno:
Thanks, come to think about it, probably makes no sense to duplicate the room. I wouldn't expect copying an event to create rooms on a conference. Not sure about tracks though.

sebix:
> Not sure about tracks though.

I'd not copy it. When copying an event I do not expect that tracks are created. The track information of the event could be copied if the same track exists also in the new conference tough.

Is this feature also available to the submitters or only to the admins?

app/controllers/events_controller.rb
```diff
@@ -220,6 +222,25 @@ def destroy
     end
   end
 
+  def copy_to_conference
+    old_event = Event.find(params[:id])
+    new_event = old_event.dup
+    authorize! :create, new_event
+
+    new_conference = Conference.find(params[:event][:conference_id])
```
artemasmith: Looks like you're missing an service object. Like `ConferenceCopier`.
```ruby
class ConferenceCopier
  attr_reader :event, :old_event
  def initialize(event)
    @old_event = event
    @event = event.dup
  end

  def call!
    event.conference = Conference.find(params[:event][:conference_id])
    event.event_attachments = old_event.event_attachments.map(&:dup)
    event.event_people = old_event.event_people.map(&:dup)
    event.save!
  end
end
```

app/controllers/events_controller.rb
```diff
@@ -139,6 +139,8 @@ def new
   def edit
     @event = Event.find(params[:id])
     authorize! :update, @event
+
+    @other_conferences = Conference.where("id != ?", @event.conference.id)
```
artemasmith: I think it should be moved to the model as a scope or to a query object.

erdgeist:
It's a little sad to see this nice feature becoming stale. I'd needed it in several instances, as well, like copying over the annual events like opening, closing, lightnings talks, etc.

But I think that the UX flow in the patch is a little awkward and I think I know how to make it better: I think that the path should always be "copy to THIS conference", because you … well obviously are organising the conference, so there's no weird filtering for the target conference.

And the only dedicated view where you're presented events from other conferences is in the people view, e.g. http://192.168.216.129:3000/en/vulcan-wolf/people/12 with a whole section "Events in other conferences". (I assumed that this naturally filters all the events you're allowed to edit, but I just noticed that you can see details about submissions in other conferences that you are not organising, which should be fixed.) In the Button list there's currently 'Show', 'Edit' and 'Destroy' but I think that 'Destroy' doesn't really belong there and might be replaced by 'Clone'.

Does that (looking for old events by user) fit your work flow, as well? Or do you actually filter by event name while switching between conferences (Which I always find annoying, because strange things happens if you hop between conferences).

I will look through implementation details once we know what we actually want.

Copying the track name in case it also exists in current conference sounds sane. Duration, and all string fields can just be duplicated, but I'd only copy over presenting EventPeople.

Maybe EventRatings should be copied over as well, in case the rating person has review/orga rights in the current conference.

PR #186, opened 2015-08-09
Add schema.org JSON-LD metadata to schedule events.

stapelberg:
This allows modern search engines and other consumers of schema.org
metadata to better understand conferences organized with frab.

The resulting metadata was successfully verified using
https://developers.google.com/structured-data/testing-tool/


manno:
I'm unsure if there is enough public and scheduled sprinkled around the json, but yes let's add this.

manno:
Merged in da056b4f

PR #188, opened 2015-08-29
Remove recording license and do not record from cfp

### LINKED ISSUE #176: Remove recording license and 'do not record' from cfp forms
Both fields should be writable by the events crew only. Putting these fields into the event submission form often collides with the information given in the actual call for participation text.

ofeldt:
This resolves #176


manno:
merged

PR #189, opened 2015-08-29
nil guard for date localization

ofeldt:
This implement a nil guard for date localization on ApplicationHelper#ldate(date, options) which will return a translated t(:date_not_set) is date is nil.

It's currently used on Event views to make them resistant to application failure if trying to be rendered with start_time attribute being nil (as in not set yet) for the event.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -41,6 +41,10 @@ module ApplicationHelper
     image_tag "icons/#{name}.png"
   end
 
+  def ldate(date, options = {})
+    [Date, DateTime, Time].include?(date.class) ? l(date, options) : t(:date_not_set)
+  end
+
   def action_button(button_type, link_name, path, options = {})
     options[:class] = "btn #{button_type}"
     if options[:hint]
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -7,11 +7,6 @@ module EventsHelper
     end
   end
 
-  def event_start_time
-    return '' unless @event.start_time
-    I18n.l(@event.start_time, format: :pretty_datetime)
-  end
-
   def timeslots
     slots = []
     (@conference.max_timeslots + 1).times do |i|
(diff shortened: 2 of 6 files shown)


app/helpers/application_helper.rb
```diff
@@ -41,6 +41,10 @@ def icon(name)
     image_tag "icons/#{name}.png"
   end
 
+  def ldate(date, options = {})
```
manno: I'd rather keep the more specific `event_start_time` method, until we have more users for a generic `ldate` method with two arguments and a somewhat cryptic name.

ofeldt: I agree on having a more descriptive name, **but** it's more a shorthand akin to `I18n.l` aliasing `I18n.localize`.
Do you want to keep the informative/verbose `I18n.t(:date_not_set)`? Otherwise rejecting this PR and making a new one by just introducting `#event_start_time` on the other view might be a simpler approach.

On a side note: Whenever i see `#model_name_attribute` on a helper i ask myself if a presenter (or viewmodel) wouldn't be more appropriate. Taking a look at `Event` itself reveals several code smells of a presenter object which tries to escape:
* `Event#humanize_time_str` - view represenation of some time attribute shouldn't be this models concern
* `Event#track_name` - reaching "through" to another object with a `#try` should be a presenters task unless we use it for internal logic
* `Event#feedback_standard_deviation` - not actually computing a number, but a string for representation purpose
* `Event#slug` - same here, this is a presenter's work.

Including `event_start_time` this gives good enough reason for such an object. One would have to take a look if  cases on other models exist, so we can pull a dependency in for doing this properly.

manno:
Commit ca0858c merges the bug fix.

The event model certainly has to many responsibilities, 282 lines ... conference and person models come to mind, too.

I'm not a big fan of draper and after introducing rails-cast-style presenters into another project I wouldn't go down that road again. Maybe it's sufficient to introduce some plain ruby objects?

I just stumbled across [Thoughts about Rails Presenters](https://gist.github.com/somebox/5a7ebf56e3236372eec4), which sums up the whole discussion nicely.

PR #192, opened 2015-09-13
Remove public flag from rooms

### LINKED ISSUE #177: Non-public rooms in schedule
Not all output formats honor the rooms public flag. See JSON and XML for a working scope.

manno:
I don't think anyone is using this, so this branch removes the public flag from rooms. This would fix #177

### DIFF
--- a/app/controllers/public/schedule_controller.rb
+++ b/app/controllers/public/schedule_controller.rb
@@ -27,7 +27,7 @@ class Public::ScheduleController < ApplicationController
     end
     @day = @conference.days[@day_index]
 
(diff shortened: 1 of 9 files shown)


zonque:
I never understood the purpose of this flag, so I'm all for removing it.

sebix:
I think we actually used it for the rooms where the LPI exams took place. They should not show up in the public schedule as then arbitrary visitors want to do the exams, watch it or make other troubles. Those who registered know the assembly point / room anyway (or just ask).

We are fine with just don't adding the LPI exams to the schedule I think. So I'm fine with it, +1

PR #194, opened 2015-10-08
change event to session

aselims:
The whole conference is an event. So a speaker should submit a "session" (a talk or a workshop) not "event" for the sake of UX! 
image


manno:
I do agree with your reasoning, but we use 'event' everywhere consistently. I don't want to change it in just one place. But to change it across the whole app would be very troublesome.

aselims:
Yeah that is what I figured out. However, those two strings I think are the main facing for users to take action.

zonque:
Well, a 'session' is what frab tracks with its cookie, so that's very misleading. Would you call any given talk on a conference a 'session' really? I take it you're a native speaker and so you rightfully have a hard opinion on this, but really, all these terms are totally overloaded anyway, aren't they?

In any case, changing the name of a model an its controller will also change the URLs, and that will certainly lead to regressions in existing installations.

manno:
We can't merge this. We'd have to keep the old names for compatibility anyways. Maybe someone could fix this in the translations.

PR #196, opened 2015-10-10
Allow any available locale to be chosen

alvarezp:
I'm not sure if this is the best way to do this, but the idea is to let languages other than "en" and "de" to be used.


manno:
I'm closing this, as all available locales are selectable now.

PR #199, opened 2015-12-11
TransportNeeds

zonque:
People attending conferences may be in need of transportation, especially if the location is not easily reachable. Depending on how the conference is ran, the organizers might be in charge of providing such a service for their speakers, for instance. We've had such a situation earlier this year. Instead of maintaining the list of such demands externally, this patch set allows tracking such information from frab.

A TransportNeed model belongs to both a person and a conference and is just yet another data type with a `has_many` association. A new reports module is added to give organizers a comprehensive overview on future transportation needs of a conference.

### DIFF
--- a/app/assets/stylesheets/frab.css
+++ b/app/assets/stylesheets/frab.css
@@ -116,6 +116,7 @@ table.room tr.even {background: #EEEEEE;}
 span.right { float: right; }
 
 td.buttons { white-space: nowrap; }
+td.nowrap { white-space: nowrap; }
 
 a.assoc {
   margin-left: 150px;
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -124,9 +124,9 @@ class ConferencesController < ApplicationController
 
   def conference_params
     params.require(:conference).permit(
-      :acronym, :title, :timezone, :timeslot_duration, :default_timeslots, :max_timeslots, :feedback_enabled, :expenses_enabled, :email,
-      :program_export_base_url, :schedule_version, :schedule_public, :color, :ticket_type, :event_state_visible, :schedule_custom_css,
-      :schedule_html_intro, :default_recording_license,
+      :acronym, :title, :timezone, :timeslot_duration, :default_timeslots, :max_timeslots, :feedback_enabled, :expenses_enabled,
+      :transport_needs_enabled, :email, :program_export_base_url, :schedule_version, :schedule_public, :color, :ticket_type,
+      :event_state_visible, :schedule_custom_css, :schedule_html_intro, :default_recording_license,
       rooms_attributes: %i(name size public rank _destroy id),
       days_attributes: %i(start_date end_date _destroy id),
       tracks_attributes: %i(name color _destroy id),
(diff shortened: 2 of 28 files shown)


manno:
I'm undecided, but having this lying around doesn't help either.
Your conference is probably the only one which needs this? Should I just merge this pr, or ask the mailing list for opinions?

zonque:
> 'm undecided, but having this lying around doesn't help either.
> Your conference is probably the only one which needs this? Should I just merge this pr, or ask the mailing list for opinions?

Hmm, we had to transport people, and as all those people were already present in the database, we figured the easiest way to do cope with this task is to attach the information about who needs a lift when to the existing data model. The alternative would have been to maintain a separate list in some sort of Wiki, which was considered more error-prone at the time.

I'm not sure whether we're the only ones who will use the feature eventually. TBH, I can't believe we're the only conference that offers their speakers and artists assistance in logistics.

Also, it doesn't really hurt to have it, does it? Even if it's not used in most installations, the worst that can happen is minimal disk space consumption due to an empty database table.

MaZderMind:
The worst thing that can happen is having unmaintained code that slowly rots away, because it's not used and tested. Despite that I'd opt for merging because I would imagine it could be useful in a C3/Camp environment, too.

manno:
I did monitor all travel arrangements in a spreadsheet and used the ticket system for communication. 
However we had no bus and shuttle services, we mostly organized flights.
Just like reimbursements I considered this out of scope and the data to private to have it in frab. 

It's not about the database size, though. I'm afraid nobody is going to maintain the code. Frab is already a large application and this feature introduces more controllers, instance variables in views, another `if` case in a view, more menu items, etc. 

But well, we could just merge it and clean up later ... add more partials, dry up the reports controller, put the queries back into the models, remove ivars from views and do whatever rubycritic and code climate recommend ;)

zonque:
> Just like reimbursements I considered this out of scope and the data to private to have it in frab.

Fair enough, I can accept that. The question then is what the boundaries of frab are, in general. I always considered it a tool that helps organize a conference, with all it takes to make that happen.

That said, I've never looked into the ticket system integration really. If that allows us to handle such information, I'm happy to change our work-flow.

zonque:
After having a look at the remote ticket implementation as it currently stands, I have now confidence it is unsuitable of tracking transportation needs of people, as it only operates on events.

However, the latter can be changed when polymorpic associations are used. I can work on that, but I fear we would need to have support for multiple tickets per person or event in order to make that more useful. And even then, keeping the frab data in sync with external issues is more than tricky. Just imagine cases where events are canceled or people disappear from the frab database for some reason, but the tickets for them around externally. This is a can of worms I am very sure should not be opened.

Any opinion?

manno:
If you need this we might as well merge it. 

frab needs a 'feature toggle' system, but it would be sufficient if travel reimbursement could be toggled per conference, just like 'event_state_visible'.

zonque:
> frab needs a 'feature toggle' system, but it would be sufficient if travel reimbursement could be toggled per conference, just like 'event_state_visible'.

That's a good idea, will work on that.

zonque:
Ok, this is now implemented. While at it, I also made the expenses feature an optional per-conference setting. Is this what you had in mind?

manno:
Sounds good.

zonque:
Thanks :)

PR #200, opened 2015-12-11
Cleanups

zonque:
Two cleanups that shouldn't result in any visible change.


zonque:
Rebased now. If this is wanted, it should get merged soon. Due to its invasive tree-wide effect, one of these patches tends to bit-rot quickly.

manno:
It's very invasive. I had another branch depending on this one to convert double quotes to single quotes :)

Not sure if we want to do this though, there might be some frab installations with customizations?

zonque:
> Not sure if we want to do this though, there might be some frab installations with customizations?

Up to you of course, but I wouldn't worry about downstream deviations too much. IMO, if people decide to not bring their chances upstream, they have to live with the consequences.

zonque:
Now rebased to current mater.

PR #210, opened 2016-02-10
Add support for Redmine tickets

zonque:
Allow users to configure a Redmine ticket server backend.
This implementation lets `ActiveResource` handle all the json/REST magic.

The 'password' field of the configuration carries an API key instead, as
this is how their REST API works, so I tweaked the help text a bit.

From `TicketsHelper::get_ticket_view_url`, use `Conference::get_ticket_module`,
as that already does all the string-to-module dance for us.

The patch also adds an `:event_url` to the hashmap that is passed to the
`create_remote_ticket` implementations, so the modules can include a URL to
the ticket that points back to the frab event.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem 'jquery-ui-rails'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.0'
 
+gem 'activeresource'
 gem 'acts_as_commentable'
 gem 'bcrypt-ruby'
 gem 'cancancan'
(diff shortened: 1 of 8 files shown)


zonque:
Btw - I believe the ticket server implementation is somewhat backwards. I worked on patches to clean that up, but I can't test it as I don't have an OTRS or RT instance running. If anybody wants to give it a try, I'd appreciate some feedback: https://github.com/zonque/frab/commits/ticket

manno:
Looks great, however I can't test it either.

zonque:
I've set up RT and ORTS servers in VMs now, and already fixed a regression in that branch. Will do some more tests, make the Ticket model ready for polymorphic associations so it can be attached to other things than just events, and then post a new PR.

Closing this one for now.

PR #211, opened 2016-02-21
Rework ticket implementation, make association polymorphic, add support for Redmine

zonque:
This PR rewrites major parts of the support for external ticketing systems, allows tickets to be attached to both events and users, and also introduces a new helper which can talk to Redmine servers.


zonque:
Rebased to current master.

PR #212, opened 2016-02-23
event: calculate available start times in event form

zonque:
In the form for events, intersect the availability for all presenting persons, and only display those as options in the `start_time` drop down menu.

In order not to break records that have a conflicting time set, the current value is special-cased and marked accordingly. This way, people can still save such an event without accidentally altering the start time.

For this to work, the SimpleForm wrapper is provided with a block, so the default Rails form helper can kick in and make use of `grouped_options_for_select()`.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -124,6 +124,7 @@ class EventsController < ApplicationController
   def new
     authorize! :crud, Event
     @event = Event.new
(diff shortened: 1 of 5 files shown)


manno:
Sounds nice, but there is this one untested, 40 line method in the controller. Couldn't that go into the event model and be split into several methods?

zonque:
Yes, good idea, thanks! Did that now, and also added a test. Please have a look.

PR #217, opened 2016-02-29
update some gems

zonque:
Most importantly, update rails to 4.2.5.2. Refer to the blog post below for more information.
    
http://weblog.rubyonrails.org/2016/2/29/Rails-4-2-5-2-4-1-14-2-3-2-22-2-have-been-released/


zonque:
Btw - this one is rather important as it fixes two CVEs (see the blog post I linked to).

manno:
Both CVEs should not apply to frab as we don't use 'render' insecurely. brakeman should catch this. However the update contains "bug fixes for the previous security releases", so users better update all installations.

Rails 4.2.6 should be out any day now.

PR #219, opened 2016-03-10
production env: hook up exception_notification

zonque:
Hook up this very handy tool for notifing admins about exceptions
in production systems. This is off by default and needs to be enabled
via ENV['EXCEPTION_EMAIL'].


manno:
merged

PR #222, opened 2016-04-25
password_controller: set user.confirmed_at when password was changed

zonque:
When the password was changed successfully, the user must have had access to
a mail sent to his mailbox, which is, technically, nothing less than a
confirmation. Hence, set the confirmed_at field in this case.

The problem here is that people who can't log in desperately tried to reset
their password, and even succeeded in that, but were then refused to login
on the next attempt.


manno:
I wish frab would use Devise (again).

PR #223, opened 2016-05-05
Reveal person's availabilities and export event's language 

erdgeist:
The halfnarp tool has gained a planning interface called fullnarp which graphically marks conflicts in speaker avaliabilites and imbalanced event language at a given time slot.

To properly work, it needs language and all speaker's availabilites. Language can be exported directly. To map availabilities, we first need to get people ids to later xref with list of availabilities. Later we scrape all person's availabilities.

### DIFF
--- a/app/controllers/availabilities_controller.rb
+++ b/app/controllers/availabilities_controller.rb
@@ -17,6 +17,15 @@ class AvailabilitiesController < ApplicationController
     redirect_to(person_url(@person), notice: 'Availibility was successfully updated.')
   end
 
+  def show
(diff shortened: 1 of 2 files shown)


manno:
I created #224 - I think that would require less requests and was a good opportunity to clean up the controller. I guess no one is using this api (well except halfnarp), so I'd drop xml rendering. One could create an `.xml.haml` view file though.

If you can live with the nested output of PR #224 I'd merge that instead.

erdgeist:
Looks great. Gekauft.

PR #224, opened 2016-05-06
Features/exports accepted with availabilities

manno:
Add test, clean up controller and use jbuilder instead, add availabilities and language as in #223


erdgeist:
Closed #223. This PR actually does the trick. Halfnarp will filter out availabilities before pushing the export to users.

manno:
@erdgeist  deployed

PR #226, opened 2016-05-08
Never paginate for MIME types with no concept of paging

erdgeist:
Currently exporting to JSON or XML will paginate leaving the requestor no way or even a path to ask for more.

Solution: Do not paginate for anything than html.


manno:
Might make sense to add the `api-pagination` gem at some point in the future, like here: https://github.com/voc/media.ccc.de/commit/1e1125ab78d4981b92e75ca897ccc015dbd56283

PR #227, opened 2016-05-09
GROUP BY without an aggregate function makes no sense

erdgeist:
This fixes the problem in my example conference where of the two people in the only event only one the first would show up in every list that uses the involved_in function, most notably people_controller.rb.

The resulting query was

SELECT DISTINCT "people".* FROM "people" INNER JOIN "event_people" ON "event_people"."person_id" = "people"."id" INNER JOIN "events" ON "events"."id" = "event_people"."event_id" INNER JOIN "conferences" ON "conferences"."id" = "events"."conference_id" WHERE "conferences"."id" = 1 GROUP BY "people.id" LIMIT 100 OFFSET 0

and yielded an incorrect result (on the sqlite3 I found myself with after following the guide from README).

Without the GROUP BY the statement yields the correct rows from people, with the DISTINCT statement ensuring what I only can guess was meant to be done with the GROUP BY.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -38,13 +38,13 @@ class Person < ActiveRecord::Base
   # validates_inclusion_of :gender, in: GENDERS, allow_nil: true
 
   scope :involved_in, ->(conference) {
-    joins(events: :conference).where("conferences.id": conference.id).group(:"people.id")
+    joins(events: :conference).where("conferences.id": conference.id)
   }
   scope :speaking_at, ->(conference) {
-    joins(events: :conference).where("conferences.id": conference.id).where("event_people.event_role": %w(speaker moderator)).where("events.state": %w(unconfirmed confirmed)).group(:"people.id")
+    joins(events: :conference).where("conferences.id": conference.id).where("event_people.event_role": %w(speaker moderator)).where("events.state": %w(unconfirmed confirmed))
   }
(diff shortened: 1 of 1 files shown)


zonque:
Could you add a test case that shows the problem you're seeing? That would also be nice to have for future regression tests.

erdgeist:
Added a test in #228, however the issue is quiet simple: GROUP BY must never be invoked without aggregator, so adding a test here does not ensure the same error class is not repeated elsewhere.

manno:
I think that `group by` needs to be replaced by `uniq`? I haven't had time to look into it, but a person might appear several times on a conference, via `event_person`.

erdgeist:
This is not necessary. RoR already does ensure that you do not get duplicate records in queries. The actual SQL statement therefore is

SELECT  DISTINCT "people".* FROM "people" INNER JOIN "event_people" ON "event_people"."person_id" = "people"."id" INNER JOIN "events" ON "events"."id" = "event_people"."event_id" INNER JOIN "conferences" ON "conferences"."id" = "events"."conference_id" WHERE "conferences"."id" = 1 LIMIT 100 OFFSET 0

which includes a DISTINCT statement on all columns from the people table, which directly translate in what you would expect from such a function in the first place.

manno:
I believe this is fixed with #229

PR #228, opened 2016-05-09
test case for the broken group by statement

erdgeist:
This test checks if the Person.speaking_at actually shows the two speakers added to the confirmed event in the new conference.

Obviously fails on master.


manno:
I believe this is fixed with #229

zonque:
Well, but the idea here was to have a test case which prevents regressions in that area in the future, so I really think this should go in.

@erdgeist, as a prove for the assumption that is was fixed in #229, could you please rebase this branch onto the current master and force-push? The added test should succeed then, and more tests are always good to have.

PR #230, opened 2016-05-10
Factory girl include cleanup

manno:
We include FactoryGirl syntax in our `test_helper.rb`, this PR shortens all factory call builder calls.


zonque:
LGTM!

manno:
Merged via command line, because this PR is not based on master...

PR #237, opened 2016-06-02
Merge users

erdgeist:
Allows you to merge persons und their users, optionally keeping the one that was updated most recently.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -175,12 +175,47 @@ class Person < ActiveRecord::Base
 
     keep, kill = kill, keep if keep_last_updated and self.updated_at < doppelgaenger.updated_at
 
+    # Merge or move user model
     if keep.user.present?
       keep.user = keep.user.merge_with(kill.user, keep_last_updated) if kill.user.present?
     else
       keep.user = kill.user
       kill.user = nil
     end
+
+    # Get list of all conferences for which keep already has set the availabilities, then only
+    # import availabilities for the others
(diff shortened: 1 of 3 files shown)


app/models/person.rb
```diff
@@ -169,6 +169,60 @@ def remote_ticket?
     ticket.present? and ticket.remote_ticket_id.present?
   end
 
+  def merge_with(doppelgaenger, keep_last_updated=false)
+    keep = self
+    kill = doppelgaenger
+
+    keep, kill = kill, keep if keep_last_updated and self.updated_at < doppelgaenger.updated_at
+
```
manno: I find the first part up to this comment confusing. Why not add a class method `def Person.combine(keep, kill)`, move all the following code there and do something like

~~~ruby
return Person.combine(doppelgaenger, self) if keep_last_updated && newer?(doppelgaenger)
Person.combine(self, doppelgaenger) 
~~~

Extracting the `updated_at` check into a method is just for documentation of course.

(I kind of regret using explicit `self` in most of frabs code, since you only need it in assignments. Most code bases omit it.)

erdgeist:
Well, this pretty much has to do with this feature's users. I assumed there would be an "advanced" tab (for frab admins only?) in a Person view, allowing to merge in another Person, thus the improvised API making it belong to the Person displayed. 

Your example code snippet I do not understand, on whom would you call newer?(), if this was a class method?

erdgeist:
Handing it off to #240

PR #238, opened 2016-06-02
Allow filtering event list for tracks, states and event_types

erdgeist:
Restrict list of displayed events to be filtered by track name, state or type allows track teams to just work on their track and still have all the search and sorting-options.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -220,9 +220,9 @@ class EventsController < ApplicationController
 
   def search(events, params)
     filter = events
-    filter = filter.where( {state: params[:state]}) if params[:state].present?
-    filter = filter.where( {event_type: params[:type]}) if params[:type].present?
-    filter = filter.where( {track: Track.find_by(:name => params[:track_name])}) if params[:track_name].present?
+    filter = filter.where(state: params[:event_state]) if params[:event_state].present?
+    filter = filter.where(event_type: params[:event_type]) if params[:event_type].present?
+    filter = filter.where(track: Track.find_by(:name => params[:track_name])) if params[:track_name].present?
 
     if params.key?(:term) and not params[:term].empty?
       term = params[:term]
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -219,6 +219,11 @@ class EventsController < ApplicationController
   end
 
   def search(events, params)
+    filter = events
+    filter = filter.where( {state: params[:state]}) if params[:state].present?
+    filter = filter.where( {event_type: params[:type]}) if params[:type].present?
+    filter = filter.where( {track: Track.find_by(:name => params[:track_name])}) if params[:track_name].present?
+
     if params.key?(:term) and not params[:term].empty?
       term = params[:term]
       sort = begin
@@ -226,7 +231,7 @@ class EventsController < ApplicationController
              rescue
                nil
              end
-      @search = events.ransack(title_cont: term,
+      @search = filter.ransack(title_cont: term,
                                description_cont: term,
                                abstract_cont: term,
                                track_name_cont: term,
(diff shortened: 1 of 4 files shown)


app/controllers/events_controller.rb
```diff
@@ -219,22 +219,27 @@ def clean_events_attributes
   end
 
   def search(events, params)
+    filter = events
+    filter = filter.where( {state: params[:state]}) if params[:state].present?
```
zonque: The curly braces are not needed, you can drop them.

app/controllers/events_controller.rb
```diff
@@ -219,22 +219,27 @@ def clean_events_attributes
...
+    filter = filter.where( {event_type: params[:type]}) if params[:type].present?
```
zonque: This doesn't work: the URL parameter is called `event_type`, but here you're checking for `type`.

zonque:
I like the idea, but the fact that the filters selected by clicking on the track name, state, etc are &'ed might not be expected by everyone. I would suggest adding some informational text next to the search bar that tells users which filters are currently active. Extra bonus points for a way to disable them one by one :)

erdgeist:
The list of active filters (if any) is now displayed above the search bar, along with an [x] 'button' to remove them one by one. If you have a better idea how to visually hint that they can be removed, I'd be glad to hear.

erdgeist:
Hmm, also the way to check if the conference actually has been empty:
`if params[:term].blank? and @events.all.empty?`
appears a little cumbersome. Why not actually test for `Events.all.empty?` instead?

zonque:
> Hmm, also the way to check if the conference actually has been empty:
> if params[:term].blank? and @events.all.empty?
> appears a little cumbersome. Why not actually test for Events.all.empty? instead?

I'm not following. Where is that code used? I can't seem to find it.

In any case, `Event.all` (uppercase) refers to *all* events in the database, but you probably only want to look at those belonging to the current conference. `@conference.events.any?` would tell your whether there are any events in the current conference, but it's hard to give further hints without knowing what you're trying to do.

erdgeist:
`app/views/events/index.html.haml` displays an informative message that no events have been added to the conference yet, using the quoted query. @events has been modified by the new filter, though, showing this alert, even if there actually **are** events.

zonque:
> app/views/events/index.html.haml displays an informative message that no events have been added to the conference yet, using the quoted query. @events has been modified by the new filter, though, showing this alert, even if there actually are events.

Ah, now I see what you mean, thanks. In that case, the correct check would be `@conference.events.any?`, ignoring `params[:term]` and `@events` entirely.

erdgeist:
Jo klappt. Thanks, committed. Feeling rather good about the pr now.

app/views/events/_filters.html.haml
```diff
@@ -0,0 +1,15 @@
...
+        = link_to "[x]", url_for(params.except(:track_name))
+        track name: #{params[:track_name]}
+    %span{:style => 'margin-left: 1em'}
+      - if params[:event_type].present?
+        = link_to "[x]", url_for(params.except(:event_type))
+        event type: #{params[:event_type]}
+    %span{:style => 'margin-left: 1em'}
+      - if params[:event_state].present?
+        = link_to "[x]", url_for(params.except(:event_state))
+        event state: #{params[:event_state]}
```
zonque: As this is a repeating pattern, such code could be moved to a helper function.

zonque:
Yeah, looks good now, except for the proposed de-duplication. But that could be done in a subsequent patch.

erdgeist:
Well, with just three occurrences and no other user in sight, I ended up saving 2 lines and adding complexity. Should we need something similar in Person filters, I think it'd be justified.

zonque:
Right. If more users emerge, this can still be refactored. Thanks for working on this, lgtm!

manno:
Rebased and merged this PR.

PR #241, opened 2016-06-06
in case of an empty track, do not display a filter link

erdgeist:
If no track is set, the link to the track filter would be broken.


zonque:
:+1:

PR #242, opened 2016-06-14
Roust gem

erdgeist:
The roust gem is a wrapper around RTs REST API.

While updating other gems to the current version, I also fixed the code to adopt changes to PaperTrail, as described here: https://github.com/airblade/paper_trail/blob/master/doc/warning_about_not_setting_whodunnit.md


Gemfile
```diff
@@ -53,6 +53,8 @@ gem 'prawn', '< 1.0'
 gem 'prawn_rails'
 gem 'ransack'
 gem 'ri_cal'
+#gem 'roust'
+gem 'roust', :git => 'git@github.com:bulletproofnetworks/roust.git'
```
zonque: Please use the `name: value` notation. I recently unified all this tree-wide, and we shouldn't introduce new users of the old-fashion syntax. Same counts for the actual code in this commit.

Gemfile.lock
```diff
@@ -114,6 +114,9 @@ GEM
       activesupport (>= 4.1.0)
     haml (4.0.7)
       tilt
+    httparty (0.13.7)
+      json (~> 1.8)
+      multi_xml (>= 0.5.2)
```
zonque: While at it, you could as well squash this commit into the other one that touches `Gemfile`.

Gemfile
```diff
@@ -53,8 +53,8 @@ gem 'prawn', '< 1.0'
 gem 'prawn_rails'
 gem 'ransack'
 gem 'ri_cal'
-#gem 'roust'
-gem 'roust', :git => 'git@github.com:bulletproofnetworks/roust.git'
+gem 'roust'
+#gem 'roust', :git => 'git@github.com:bulletproofnetworks/roust.git'
```
zonque: `gem 'roust', github: 'bulletproofnetworks/roust'` should work.

zonque:
Don't know why GitHub allowed me to comment on outdated commits, but the comments still all apply, as far I can tell.

sebix:
> Don't know why GitHub allowed me to comment on outdated commits, but the comments still all apply, as far I can tell.

Why shouldn't you be to allowed to comment on any commits? *outdated* does mean that the line has been affected by a newer commit. I don't see a reason which justifies disablement of comments.

zonque:
> Why shouldn't you be to allowed to comment on any commits? outdated does mean that the line has been affected by a newer commit. I don't see a reason which justifies disablement of comments.

Sure, I know that. I just wonder why GitHub presented the commits I commented on as the most recent ones.

sebix:
I guess you commented on the particular commits themselves, not the changes-overview of the pull request?

PR #243, opened 2016-06-14
Add code to return to the url requested before login 302

erdgeist:
if users are redirected to the login page, add a helper to redirect them to their previous url when successfully logged in.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,7 +94,11 @@ class ApplicationController < ActionController::Base
     if request.path =~ /\/cfp/
       new_cfp_session_path
     else
-      new_session_path
+      if request.get?
+        new_session_path(:return_to => request.path)
+      else
+        new_session_path
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -94,7 +94,11 @@ def scoped_sign_in_path
     if request.path =~ /\/cfp/
       new_cfp_session_path
     else
-      new_session_path
+      if request.get?
+        new_session_path(:return_to => request.path)
```
zonque: `return_to: request.path` looks much nicer and is also what the code uses elsewhere.

zonque:
Just curious: Is there any particular reason why you chose not to store this information in the cookie?

erdgeist:
Because we usually do have a bunch of tabs open that belong to the same session and I do have no clue how to properly separate information stored in the cookie. I am afraid all tabs end up on the same view after login.

manno:
Was wondering if it's ok to redirect to a user controlled value, but Rails no longer accepts a hash for redirect_to and the login method is not well suited for URL spoofing attacks.

PR #256, opened 2016-07-26
FrOSCon Public Schedule Style

manno:
Salvaged from last years FrOSCon program at: https://programm.froscon.org/2015/events.html


manno:
This would need some clean up, but I like the way it handles more than seven rooms. Rendering them in a new table below the first one. The current solution is broken and we could get rid of https://github.com/frab/frab/blob/master/app/assets/javascripts/public_schedule.js.coffee

This also gets renames the last 'horizontal-navigation' classes, which were unused (?).

Rendering the tracks into the side navi, instead of creating an additional top navi is great, too.

Listing the events by track on the events page is debatable.

PR #261, opened 2016-09-19
Add support for sub-conferences

zonque:
This PR adds support for sub-conferences, so that parts of a conference can be delegated to a separate team to handle their own events, while administrators of the 'master' conference stay in control of specific details.

The implementation details are:

* Each conference gains a `parent` field to link to another conference. During the creation of a conference, all existing conferences the currently logged in user has administration rights for may be selected as parent.

* A sub-conference inherits all `days` from `parent` and may not create any `day` itself.

* The `timezone` and `timeslot_duration` attributes in sub-conferences are locked and hidden from the forms. The values are taken from `parent` instead.

* Rooms and tracks of a sub-conference may only be edited by users that have administration rights for the parent conference.

* In the public schedule of the parent conference, all events, tracks, rooms etc from all sub-conferences are taken into account for the HTML view and all machine-readable output formats. However, they don't appear in the usual work-flow in the parent conference.

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -30,6 +30,7 @@ class ConferencesController < ApplicationController
   def new
     params.delete(:conference_acronym)
     @conference = Conference.new
+    @possible_parents = Conference.where(parent: nil)
     @first = true if Conference.count == 0
 
     respond_to do |format|
@@ -50,6 +51,10 @@ class ConferencesController < ApplicationController
   # POST /conferences
   def create
     @conference = Conference.new(conference_params)
+
+    if @conference.parent and not can? :administate, @conference.parent
+      @conference.parent = nil
+    end
 
     respond_to do |format|
       if @conference.save
@@ -122,16 +127,33 @@ class ConferencesController < ApplicationController
   end
 
   def conference_params
-    params.require(:conference).permit(
-      :acronym, :title, :timezone, :timeslot_duration, :default_timeslots, :max_timeslots, :feedback_enabled, :expenses_enabled,
+    allowed = [
+      :acronym, :title, :default_timeslots, :max_timeslots, :feedback_enabled, :expenses_enabled,
       :transport_needs_enabled, :email, :program_export_base_url, :schedule_version, :schedule_public, :color, :ticket_type,
       :event_state_visible, :schedule_custom_css, :schedule_html_intro, :default_recording_license,
-      rooms_attributes: %i(name size public rank _destroy id),
-      days_attributes: %i(start_date end_date _destroy id),
-      tracks_attributes: %i(name color _destroy id),
       languages_attributes: %i(language_id code _destroy id),
       ticket_server_attributes: %i(url user password queue _destroy id),
       notifications_attributes: %i(id locale accept_subject accept_body reject_subject reject_body _destroy)
-    )
+    ]
+
+    if @conference && @conference.new_record?
+      allowed += [ :parent_id ]
+    end
+
+    if @conference && @conference.parent.nil?
+      allowed += [
+        :timezone, :timeslot_duration,
+        days_attributes: %i(start_date end_date _destroy id),
+      ]
+    end
+
+    if @conference && (@conference.parent.nil? || can?(:adminstrate, @conference.parent))
+      allowed += [
+        rooms_attributes: %i(name size public rank _destroy id),
+        tracks_attributes: %i(name color _destroy id),
+      ]
+    end
+
+    params.require(:conference).permit(allowed)
   end
 end
--- a/app/controllers/public/schedule_controller.rb
+++ b/app/controllers/public/schedule_controller.rb
@@ -27,7 +27,7 @@ class Public::ScheduleController < ApplicationController
     end
     @day = @conference.days[@day_index]
 
-    @all_rooms = @conference.rooms
+    @all_rooms = @conference.rooms_including_subs
     @rooms = []
     @events = {}
     @skip_row = {}
@@ -52,7 +52,7 @@ class Public::ScheduleController < ApplicationController
   end
 
   def events
-    @events = @conference.events.is_public.confirmed.scheduled.sort { |a, b|
+    @events = @conference.events_including_subs.is_public.confirmed.scheduled.sort { |a, b|
       a.to_sortable <=> b.to_sortable
     }
     @events_by_track = @events.group_by(&:track_id)
@@ -64,8 +64,8 @@ class Public::ScheduleController < ApplicationController
   end
 
   def event
-    @event = @conference.events.is_public.confirmed.scheduled.find(params[:id])
-    @concurrent_events = @conference.events.is_public.confirmed.scheduled.where(start_time: @event.start_time)
+    @event = @conference.events_including_subs.is_public.confirmed.scheduled.find(params[:id])
+    @concurrent_events = @conference.events_including_subs.is_public.confirmed.scheduled.where(start_time: @event.start_time)
     respond_to do |format|
       format.html
       format.ics
@@ -73,7 +73,7 @@ class Public::ScheduleController < ApplicationController
   end
 
   def speakers
-    @speakers = Person.publicly_speaking_at(@conference).confirmed(@conference).order(:public_name, :first_name, :last_name)
+    @speakers = Person.publicly_speaking_at(@conference.include_subs).confirmed(@conference.include_subs).order(:public_name, :first_name, :last_name)
     respond_to do |format|
       format.html
       format.json
@@ -82,7 +82,7 @@ class Public::ScheduleController < ApplicationController
   end
 
   def speaker
-    @speaker = Person.publicly_speaking_at(@conference).confirmed(@conference).find(params[:id])
+    @speaker = Person.publicly_speaking_at(@conference.include_subs).confirmed(@conference.include_subs).find(params[:id])
   end
 
   def qrcode
--- a/app/helpers/public/schedule_helper.rb
+++ b/app/helpers/public/schedule_helper.rb
@@ -24,7 +24,7 @@ module Public::ScheduleHelper
   end
 
   def different_track_colors?
-    colors = @conference.tracks.map(&:color)
+    colors = @conference.tracks_including_subs.map(&:color)
     colors.uniq.size > 1
   end
 
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -12,8 +12,10 @@ class Conference < ActiveRecord::Base
   has_many :conference_exports, dependent: :destroy
   has_many :mail_templates, dependent: :destroy
   has_many :transport_needs, dependent: :destroy
+  has_many :sub_conferences, class_name: Conference, foreign_key: :parent_id
   has_one :call_for_participation, dependent: :destroy
   has_one :ticket_server, dependent: :destroy
+  belongs_to :parent, class_name: Conference
 
   accepts_nested_attributes_for :rooms, reject_if: proc { |r| r['name'].blank? }, allow_destroy: true
   accepts_nested_attributes_for :days, reject_if: :all_blank, allow_destroy: true
@@ -35,6 +37,8 @@ class Conference < ActiveRecord::Base
   validates :acronym, format: { with: /\A[a-zA-Z0-9_-]*\z/ }
   validates :color, format: { with: /\A[a-zA-Z0-9]*\z/ }
   validate :days_do_not_overlap
+  validate :subs_dont_allow_days
+  validate :subs_cant_have_subs
 
   after_update :update_timeslots
 
@@ -55,6 +59,32 @@ class Conference < ActiveRecord::Base
   scope :accessible_by_orga, ->(user) {
     joins(:conference_users).where(conference_users: { user_id: user, role: 'orga' })
   }
+
+  alias :own_days :days
+
+  def days
+    if parent
+      parent.days
+    else
+      own_days
+    end
+  end
+
+  def timezone
+    if parent
+      parent.timezone
+    else
+      self.attributes['timezone']
+    end
+  end
+
+  def timeslot_duration
+    if parent
+      parent.timeslot_duration
+    else
+      self.attributes['timeslot_duration']
+    end
+  end
 
   def self.current
     self.order('created_at DESC').first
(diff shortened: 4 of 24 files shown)


manno:
Thanks for trying to create this in a hurry. While there is some confusion about this feature, it seems it is desperately needed for the 33c3.

While this is an interesting prototype I don't think this PR is usable in its current state.
Since there is a typo in the first changed line ("adminstrate") I guess this hasn't been tested much.

## Design

In previous discussions we considered two ways to introduce sub conferences:
* Extend ability to work on tracks, sub conferences are just tracks of a conference
* Introduce a class hierarchy to have properly modeled main/sub conference classes and relations to other models

This approach adds a self relation to the conference model and overrides methods to point to either the parent or subs relations, i.e. on rooms. While this looks easy at first, it's really just hiding the sub/main conference relation. Instead of explicit types switch statements are introduced.

While the change is relatively small, I don't believe it works. Sadly the current test suite  is not big enough to catch this and this PR does not introduce enough tests to ensure core functionality is still working.

If you want to continue on this implementation, please try to run the existing tests with a 'main conference' and with a 'child conference'. 

I tried to do this by modifying the conference factory in `test/factories.rb`. Something like this might get you started, though I don't think I got the first one right:

### Child Conference

    diff --git a/test/factories.rb b/test/factories.rb
    index 94fac6d..c3a25e1 100644
    --- a/test/factories.rb
    +++ b/test/factories.rb
    @@ -138,6 +138,15 @@ FactoryGirl.define do
        transport_needs_enabled true
        schedule_public true
        timezone 'Berlin'
    +   parent_id nil
    +
    +   transient do
    +     build_parent true
    +   end
    +
    +   after(:create) do |c, evaluator|
    +     c.parent_id = create(:three_day_conference, build_parent: false).id if evaluator.build_parent
    +   end

        factory :three_day_conference, traits: [:three_days]
        factory :three_day_conference_with_events, traits: [:three_days, :with_rooms, :with_events]

### Parent Conference

    diff --git a/test/factories.rb b/test/factories.rb
    index 94fac6d..09dbdea 100644
    --- a/test/factories.rb
    +++ b/test/factories.rb
    @@ -138,6 +138,11 @@ FactoryGirl.define do
        transport_needs_enabled true
        schedule_public true
        timezone 'Berlin'
    +   parent_id nil
    +
    +   after(:create) do |c, evaluator|
    +     create(:conference, parent_id: c.id) if c.parent_id.nil?
    +   end

        factory :three_day_conference, traits: [:three_days]
        factory :three_day_conference_with_events, traits: [:three_days, :with_rooms, :with_events]

## Questions

* How are speakers time conflicts handled, speakers might be active in several sub-conferences? 
* How are rooms separated across sub conferences, can every conference just use every room?
* How are event conflicts handled if the conference and sub-conferences share the same rooms?

## Code Style

Here is a list of some issues I have with the code style. 

* Consider rubocop warnings
* Avoid logic in views
* Avoid ternary operator. I know the [Ruby style guide](https://github.com/bbatsov/ruby-style-guide) thinks it's ok, but no.
* Avoid nesting if, return early
* Avoid self.attributes[:aa], use self[:attribute] if possible
* Avoid += [{}] .. with hashes inside? Isn't this just << {} ?
* Extract complex if conditions to method to improve readability
* Hidden class, consider polymorphism instead of switch statement

zonque:
Thanks for having a look. I agree there more tests are needed, but I disagree there is a real risk of breaking currently existing "core functionality".

> In previous discussions we considered two ways to introduce sub conferences:
> 
>     Extend ability to work on tracks, sub conferences are just tracks of a conference
>     Introduce a class hierarchy to have properly modeled main/sub conference classes and relations to other models

I believe these are a much more complicated approaches which arbitrarily limit the functionality. Why not treat sub-conferences as normal conferences, and inherit all the benefits that come with that? For instance, there is no reason why a sub-conference shouldn't be able to export its own public schedule, reach out their speakers, have access to reports, track expenses and transport needs etc.

The only thing that is needed for such a sub-conference is that its data records appear in the public schedule of the parent conference. And as that has influence on the layout etc, only the admin of that parent conference should be able to edit tracks and rooms, which is what my PR implements.

> This approach adds a self relation to the conference model and overrides methods to point to either the parent or subs relations, i.e. on rooms. While this looks easy at first, it's really just hiding the sub/main conference relation.

Yes, it makes the change where it is needed. All other details of a conference can stay the same. Not sure why you dislike the simplicity here.

In fact, it would even be possible to nest conferences in arbitrary depth. The code currently forbids that explicitly (`.where(parent: nil)`) but it would just work if that check is removed.

> Instead of explicit types switch statements are introduced.

Where? With sub-conferences as distinct object class you would probably need such a switch statement, but I didn't introduce any of that to my knowledge. What I introduced, however, are helper functions to access data from both the conference and all sub-conferences as well.

> Since there is a typo in the first changed line ("adminstrate") I guess this hasn't been tested much.

True, I added this check last minute, but it turns out it's not even needed. As abilities are currently set up, only administrators may create new conferences anyway, so the check can go away. I just made this change.

For the rest of your comments, I'd much prefer if you could put them inline to the actual hunks, for discussing them on context. Will try to answers them inline for now.

> How are speakers time conflicts handled, speakers might be active in several sub-conferences?

This is indeed a missing bit, yes. I can look into that.

> How are rooms separated across sub conferences, can every conference just use every room?

They aren't. Sub-conferences may have as many rooms as they want, just like any other conference. But as they appear in the public schedule of the master conference, editing them is a privileged operation that can only be conducted by a user that has permission to alter the master conference rooms.

The only draw-back I can see with this approach is that a room is exclusive to a (sub-) conference. IOW, a room cannot be shared between multiple conferences, but I guess that okay.

> How are event conflicts handled if the conference and sub-conferences share the same rooms?

They don't. That's part of the design choice. And I believe that matches the requirements of most conferences - if there is a sub-conference, it is usually physically separated. 

>     Consider rubocop warnings

Will have a look.

>     Avoid logic in views

Where exactly did I introduce that? All the changes I introduced make distinctions based on whether or not the parent is present, but that's completely okay in views. But maybe this becomes clear once you put your comments inline.

>     Avoid ternary operator. I know the Ruby style guide thinks it's ok, but no.

That's a matter of taste, but if you dislike them, I can change them around.

>     Avoid nesting if, return early

Where?

>     Avoid self.attributes[:aa], use self[:attribute] if possible

It's not possible, no. When overriding the default rails getter methods, you have to fall back to `self.attributes[:aa]`.

>     Avoid += [{}] .. with hashes inside? Isn't this just << {} ?

If you are referring to the change in `conferences_controller.rb`, this is how the rails param logic works - it uses an array with either symbols or hashes, while the latter defines with nested attributes are allowed.

>     Extract complex if conditions to method to improve readability

Where? Please comment inline to the patch.

>     Hidden class, consider polymorphism instead of switch statement

Same here. I fail to see a hidden class?

zonque:
> Consider rubocop warnings
> Avoid ternary operator. I know the Ruby style guide thinks it's ok, but no.

FWIW, these are addressed now.

zonque:
The commit stack grew by one which adds new traits to run the existing tests with sub-conferences, and at least some of them with a parent conference. More details in the commit log of that new patch.

manno:
Hi,
Thanks for keeping up with this. I played around with the PR some more and have more questions.

Since deep nesting of conferences is not  supported by the queries, can we have a model validation to prevent this?

Sub Conferences can't have days, there should be a validation in the model to enforce this.

What should happen if an existing conference gets a parent. Can the days be deleted? Or do we just keep them in the database. What happens to existing, scheduled, events?

The public schedule of my main conference only renders rooms and events of the child conference. Might be my setup, though.

zonque:
> Since deep nesting of conferences is not supported by the queries, can we have a model validation to prevent this?

Good idea, yes. Added them now.

> Sub Conferences can't have days, there should be a validation in the model to enforce this.

Yes, easy enough. Added now.

> What should happen if an existing conference gets a parent. Can the days be deleted? Or do we just keep them in the database. What happens to existing, scheduled, events?

I thought about this as well, and figured it is not anything that must be supported. That's why the parent selection field is only shown in the form for a new conference, and not if the conference is edited. The controller also doesn't permit the parameter to be set, and now even the models have a check to prevent that. So we should be safe.

> The public schedule of my main conference only renders rooms and events of the child conference. Might be my setup, though.

Hmm, that's strange. That should work, and it did in my tests. Care to check your setup again?

Thanks!

zonque:
> The public schedule of my main conference only renders rooms and events of the child conference. Might be my setup, though.

Btw - there is now a new controller-level test that asserts the output of the json output.

erdgeist:
I wonder how 33C3 will benefit from any of this. Maybe I can not see the use cases yet, but even the most simple things won't work:

I would get it, if you could limit a sub-con team to a certain room (i.e. party floor) and let the party orga do their booking there. Or have assemblies work out their own schedule. But this is not how it works, right? Every team has control over every room and the party schedule must fit the time slot configuration of the main event and can not do stuff outside the parent con's lecture hours? This clearly has not been thought through from an end user's perspective.

33C3 would need a concept of track teams, to limit rights seeing reviewers notes on lectures of other tracks. Events will belong to the sub cons but not the parent con? And in order to work together on the parent con (i.e. building the final Fahrplan, discussing overall time budget constraints), all track teams would need rights to look into other sub-cons if they want to get basic info as visible to reviewers? So you'd need to grant each team reviewer permissions in all the other sub-cons or just the parent con, i.e. are the permissions transitive? Would I need to grant all reviewers permissions in all sub cons? If it's the other way around, can I limit reviewer's permissions so that party orga has no interferences with lecture track reviewers?

From what I understand you can't move events between (sub)-cons? I am constantly getting requests to make the inter-track-divide more fluid, i.e. have tracks rather be tags, and this approach means, they're even more formally separated into different conferences. So if we're using sub-cons to separate lecture tracks or if we use them to separate main con from assemblies/party we need a robust way to move events between sub-cons (but obviously not between cons).

Can person's availabilities differ for sub-cons? I wonder how conflict resolution will work.

I other words: I'm missing this feature's rationale. I do not get the vision what the problems are that this approach would solved.

sebix:
For the last Linuxwochen Wien 2016 we created a new temporary conference for the sub-conference Minidebconf16. After the schedule was final, we moved all events from one conference to the other, putting them all into one track (in order to differentiate them) and room by directly manipulating the database.

Both, the moving via database, and the track are not good solutions. The team then has no chance to change something and the subconf can't have tracks :(

Linuzifer:
@erdgeist I sense a fundamental misunderstanding here.

The Feature is intended to implement _subconferences_, not track teams.

> I would get it, if you could limit a sub-con team to a certain room (i.e. party floor) and let the party orga do their booking there. Or have assemblies work out their own schedule. But this is not how it works, right?

This is exactly what this feature is about.

33C3, like many other conferences, has independent groups, contributers and assemblies that offer their own stages and schedule. As of today, they are hidden in (mostly offline) wikis or their own sites, thus not centrally accessible in the Fahrplan or Fahrplan-based apps.

I see a great benefit for conference attendees here.

zonque:
> I wonder how 33C3 will benefit from any of this.

Note that this PR is in no way limited to 33C3. Neither is the entire project.

> I would get it, if you could limit a sub-con team to a certain room (i.e. party floor) and let the party orga do their booking there. Or have assemblies work out their own schedule. But this is not how it works, right? Every team has control over every room and the party schedule must fit the time slot configuration of the main event and can not do stuff outside the parent con's lecture hours?

No, every conference has their own rooms and own tracks. The details that are inherited from the parent conference are the days the conference takes place (obviously), the timezone (as sub-conferences are supposed to be co-located to the parent conference, that's kind of logical too) and the time slot length (internal data model specifics make this necessary). Every organization member of a sub conference may put events to their own schedule, in one of their own rooms and tracks.

There are two things that may only be edited by an administrator of the parent conference: tracks and rooms. I chose to limit access to them due to the fact that they are displayed int he parent's public schedule.

> This clearly has not been thought through from an end user's perspective.

These are strong words, especially in the light of your expressed uncertainties about the implementation details. I'd really appreciate if you checked out the PR and played around with the code. Guessing what the patches might or might not do is not helpful at this stage.

> 33C3 would need a concept of track teams, to limit rights seeing reviewers notes on lectures of other tracks. Events will belong to the sub cons but not the parent con? And in order to work together on the parent con (i.e. building the final Fahrplan, discussing overall time budget constraints), all track teams would need rights to look into other sub-cons if they want to get basic info as visible to reviewers? So you'd need to grant each team reviewer permissions in all the other sub-cons or just the parent con, i.e. are the permissions transitive? Would I need to grant all reviewers permissions in all sub cons? If it's the other way around, can I limit reviewer's permissions so that party orga has no interferences with lecture track reviewers?

I don't have a full understanding about what you guys have in mind for your specifics at 33C3, but what you are describing is clearly out of scope of this PR. I never meant to implement any of that either. What this PR implements instead is a much simpler solution.

> From what I understand you can't move events between (sub)-cons? I am constantly getting requests to make the inter-track-divide more fluid, i.e. have tracks rather be tags, and this approach means, they're even more formally separated into different conferences. So if we're using sub-cons to separate lecture tracks or if we use them to separate main con from assemblies/party we need a robust way to move events between sub-cons (but obviously not between cons).

No, events are not meant to be moved between conferences. That could be added though, and if it would, other use-cases would benefit from it as well. It could also be restricted to moving events from a sub-conference to the parent, if that is what's needed. But again, this is out of scope for now.

> Can person's availabilities differ for sub-cons? I wonder how conflict resolution will work.

Yes they can. Just as they can for totally separate conferences as well. The question is rather whether speaker conflicts between events in any sub-conference can be handled nicely, but I consider that a minor detail which can (but does need to be) addressed by additional patches. Without such a feature, speakers would need to handle their appearance in multiple conflicting conferences themselves anyway, which is completely okay IMO.

> I other words: I'm missing this feature's rationale. I do not get the vision what the problems are that this approach would solved.

If there is any unclarity left that I can help sort out, please let me know.

saerdnaer:
In my view, the mention of 33C3 by @manno did not refer track teams, but to events outside the four large lecture halls. For the 33C3 this PR solves only part of the problem - namely rooms where events are moderated,  e.g. the booking of party floor or the program on stage at Sendezentrum. At 33C3 we have also unmoderated rooms where so called "self-organized sessions" (SOS) take place. When a SOS is created or modified the event ends up directly in the timetable without further examination. But this is a separate discussion and feature and not part of this PR. 

For further information on a possible data structure for C3 see https://orga.cccv.de/doku/system#beschreibung (ask me if you need the credentials to access this wiki)


> As of today, they are hidden in (mostly offline) wikis or their own sites, thus not centrally accessible in the Fahrplan or Fahrplan-based apps.
Actually that is not fully true. At the 32C3 there was a working schedlue.xml + .json with about 10-12 rooms, which was also used by an iOS and an Andorid app.

manno:
Merged in 6271e6a

PR #265, opened 2016-09-19
Use a random password when seeding the database in production mode

zonque:
Default passwords suck, and administrators of production machines tend to
forget to change them after the initial setup.

Use a random password when seeding the database in production
environment, but stick to 'test123' for development mode. Also print
the chosen password to the console, so people can actually log in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ To start frab in the production environment run
 
     RACK_ENV=production bundle rails s
 
+Note that when seeding the database in production mode, the password for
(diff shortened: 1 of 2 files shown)


Suleman-Elahi:
Can you please show me where it printed the default password ?I can't find it, I am not used to Docker.

My console log(used docker-compose): https://paste.ee/p/Im3Fg

eladeyal-intel:
Check out the logs inside db-app. It should start with precompiling assets and building the database. At this stage it reports admin password.

PR #268, opened 2016-09-27
Bulk notification refactoring

erdgeist:
This PR contains several changes to allow sending bulk notifications to all rejected and accepted users at once and to properly inform speakers of date and location of their lecture.

A new schedule notification has been added and the edit_notification view now has buttons to issue bulk notification runs for the event state transitions that are associated with notifications:

New event states were introduced: rejecting, accepting and scheduled. rejecting and accepting are new sub-states, entered when an event is accepted/rejected and the conference's bulk-notification-flag is on. With a new notify transition the events progress to their rejected/unconfirmed state es before and now also from confirmed to scheduled, in all cases issuing a notification via RT or integrated service, depending on conference configuration.

Notification templates were re-used from the bulk-mail sub-system and allow default notification per conference, notifiy-transition and locale. Notification variable substitution has been extended to also replace %{date}, %{time} and %{room} and is now a member of EventPerson. This means that EventPersons can also receive a custom notification, that you can attach from event view with a "custom notification" button and edit from the edit_person view. This customized notification is then substituted and sent out instead of the conference-wide default. 

The ticket_server adapter now has a add_correspondence stub implemented by the rt adapter, that sends out messages to all ticket requestors or to a list of email addresses that will be temporarily made requestors, so that individualised notifications per EventPerson can be send (with proper names, confirmation tokens, etc).

The bulk notification view provides a quick overview over problems in notifiable events like missing ticket ids, missing speakers and missing notifications for the locales in current conference.

Tests and all code that relied on the former event states (i.e. code that tested for accepted lectures to be in [:unconfirmed, :confirmed] now needs to check for [:accepting, :unconfirmed, :confirmed, :scheduled]) has been brought up to date and all checks pass.


manno:
Rebased and merged in 3e0bb57771c582ef962a48aa25ec2c092972812d

manno:
Just came across the 'accepted' state. Since it's no longer part of the state machine in Event, do we need a migration for existing conferences to change all 'accepted' into 'accepting'?

erdgeist:
From my understanding "accepted" is not a state but a scope that includes everything from ["accepting", "unconfirmed", "confirmed", "scheduled"].

I can not find where "accepted" has ever been a valid state in state machine.

manno:
Thanks. I already fixed the two occurrences in https://github.com/frab/frab/commit/bdae0c7d6577fe6d7bf638d401044ab38b1167d5

wvholst:
It would be nice to extend this. For whatever reason, a lot of speakers come up with expenses requests *after* they have been accepted. It would be good for Frab to allow for mass mailings to speakers whose proposals haven't been accepted yet. So with status 'new'.

PR #274, opened 2016-10-11
Fix locale lookup broken by update in nested form

erdgeist:
The DIV in nested fields gem seems to have lost the singular class name we used to identify the parent DIV containing the SELECT for locale. This broke copying default notification texts.


zonque:
Looks good to me, thanks!

PR #280, opened 2016-10-25
Label no uppercase

erdgeist:
The informational text about what variables can be substituted is forced uppercase for being a label. But the uppercase version of variable names is not actually substituted.

This patch forces the info text to have no text-transform.


app/views/conferences/edit_notifications.html.haml
```diff
@@ -17,7 +17,7 @@
   %h1 Edit
 
   .row
-    .span6.offset8.label
+    .span6.offset8.label{ :style => "text-transform: none !important" }
```
erdgeist: I think that the style is really specific to this particular element and I wondered why bootstrap would force a text-transform in the first place. I also wanted to make sure that no more specific rule coming from another unexpected source would ever mess with the display in this element where the actual representation of the content is crucial.

Having said this, I generally agree with your remark and if you could point me to where to fix this in an acceptable way, I'd love to implement it.

erdgeist: So I noticed that the .label class was only used for background color. I added a .info-block css style and used that instead of the bootstrap .label class.

zonque:
There are now two spurious changes in environment files which shouldn't be there. Also, the change should be contained in just one commit. Could you rebase that? The actual change now looks good to me. Thanks!

erdgeist:
Yes stash gone bade, will delete this PR and open a new one.

PR #300, opened 2017-03-10
fix exception when looking at user after db:setup

### LINKED ISSUE #299: exception when looking at user after db:setup
After a fresh install, if you got to '/en/people/1/user', frab will throw an exception because @conference is not defined at that stage.

zabarabab:
fixes #299


manno:
I believe this error no longer exists. Feel free to re-open.

PR #310, opened 2017-04-10
Use https url

raoulb:
Note: This is untested but might resolve the warning.


manno:
`bundle config github.https true` might be necessary in the future.

PR #317, opened 2017-06-11
day: delete availabilities when destroyed

zonque:
When a day is destroyed, associated availabilities must be destroyed as
well. Otherwise availability.day is unresolvable afterwards.

### DIFF
--- a/app/models/day.rb
+++ b/app/models/day.rb
@@ -2,7 +2,7 @@ class Day < ApplicationRecord
   include HumanizedDateRange
 
   belongs_to :conference
-  has_many :availabilities
+  has_many :availabilities, dependent: :destroy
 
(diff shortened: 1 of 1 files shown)


manno:
That's right. They only get removed if the conference is destroyed. I wonder if it's intentionally designed like this, so the data is still there if orga makes an error. They wouldn't have to contact speakers than, but could fix it in the database.
On the other hand, database backups would be the proper solution for that.

zonque:
Sure, you could say that about any destructive database operation. As the code currently stands, a legitimate operation leads to data inconsistencies and raised exceptions.

zonque:
An alternative would be to handle the error on the caller side, and expect `availability.day` to be `nil`. Not sure if it's worth it though :)

PR #318, opened 2017-06-11
schedule: pdf export enhancement, CORS for public controller

zonque:
Two small patches: one adds some more parameters to the custom pdf creator, the other allows accessing the public schedule from any origin.


zonque:
This has already been implemented now, closing,

PR #319, opened 2017-06-11
mail templates: add a filter for speakers of all scheduled events

zonque:
With the new event state in place, this filter is needed to contact all
people who are about to speak.


zonque:
rebased now

PR #320, opened 2017-06-11
cfp/people: show event date and room in overview table

zonque:
This way, speakers can see where they are scheduled at and when.

### DIFF
--- a/app/views/cfp/people/_table.html.haml
+++ b/app/views/cfp/people/_table.html.haml
@@ -5,6 +5,7 @@
       %th= t("activerecord.attributes.event.title")
       - if @conference.event_state_visible
         %th= t("activerecord.attributes.event.state")
+        %th= t("cfp.scheduled")
       - if @conference.feedback_enabled
         %th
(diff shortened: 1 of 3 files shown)


zonque:
Rebased to trigger the tests

zonque:
To give some more background: We've had the situation that speakers would need to search for themselves in the public schedule once it's published, or come up at the information desk to ask for that. This patch helped a lot.

manno:
I'm hesitant because some conference might not want speakers to now directly. It's a secret if the talk has been accepted until the confirmation mails are sent: https://github.com/frab/frab/blob/master/app/views/cfp/people/_table.html.haml#L17
Location and time are communicated by coordinators via mail, because they might change a lot until the public schedule is finalized.

I agree it would be great to have it in the interface, but we should gate it by checking `conference.event_state_visible`.

zonque:
Good point! Will change accordingly.

manno:
fixed indentation and merged

PR #322, opened 2017-06-13
conference: additionally order by id in ::current

zonque:
When creating multiple conference in a batch operation, chances are they
all have the same created_by attribute.

This causes the default ordering scope to return the wrong conference
through Conference.current which happens in the test cases.

Fix this by ordering by their id as well.

### DIFF
--- a/app/models/conference.rb
+++ b/app/models/conference.rb
@@ -64,7 +64,7 @@ class Conference < ActiveRecord::Base
   }
 
   def self.current
-    order('created_at DESC').first
+    order('created_at DESC, id DESC').first
   end
 
   alias own_days days


zonque:
Interesting. This triggers other bugs in the test suite, but the one at hand seems to be fixed.

manno:
Sorry it will be some time before I can look at your pull requests and port them. I'm currently working on a huge refactoring regarding the switch from cancancan to pundit. If you're interested in the latest development branch take a look at the Rails 5 pf/devise branch.  I hope to beta-test all of that on Froscon in two weeks, though there is a staging system too.

zonque:
Sure, no hurries. I was recently involved in a conference using frab, and these few patches resulted from it. I just wanted to park them somewhere. Let me know if I need to rebase anything after your rework.

zonque:
This commit was cherry-picked as e7abb19c. Closing.

PR #323, opened 2017-06-14
Track: add relation to events

zonque:
Even though this is not used in the code, it's handy to have in the console.


zonque:
Rebased now.

PR #330, opened 2017-08-19
Add docker support

lanrat:
Adding initial support for running frab inside of a docker container.


manno:
Hi, Thanks for working on this. I have yet to try it out. This basically starts up containers for a production frab with puma and postgresql?

What's everyones opinion about this, should this go into the main repo? I think it's very useful?

lanrat:
By default the container starts in production mode with puma and sqlite3. However all of the environment variables and settings can be overridden. For example, the provided example `docker-compose.yml` file overrides some defaults and sets the database to use postgresql. Starting in development mode is as simple as setting the correct environment variable.

manno:
Works great, thanks!

lanrat:
Thanks for merging this!

Would it also be possible for frab to setup an account on the [Docker Hub](https://hub.docker.com/) and enable an [automated build](https://docs.docker.com/docker-hub/github/) so that whenever an update is pushed to frab the container is also updated?

Thanks.

(I happy to create the account and maintain the build for frab, but I will need the ability to add a hook to the github repo.)

manno:
Thanks, good idea. I created this https://hub.docker.com/r/frab/frab/

lanrat:
Awesome!

I created a pull request #345 that documents this as well.

PR #334, opened 2017-08-20
Add kqueue support for BSD systems

erdgeist:
I do not know if the require 'rbconfig' in Gemfile is correct, but rails keeps nagging me to use kqueue under *bsd and I whole-heartedly agree that polling on BSDs is insane.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,12 @@ git_source(:github) do |repo_name|
   "https://github.com/#{repo_name}.git"
 end
 
+require 'rbconfig'
+if RbConfig::CONFIG['target_os'] =~ /(?i-mx:bsd|dragonfly)/
+  gem 'rb-kqueue', '>= 0.2'
+end
+
+
 if ENV['CUSTOM_RUBY_VERSION']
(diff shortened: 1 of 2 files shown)


manno:
I have no idea how to test the consequences of this change. I guess it's harmless?

erdgeist:
This is what I usually am educated by, when starting the rails app:

```
=> Rails 5.1.2 application starting in development on http://192.168.216.129:3000
=> Run `rails server -h` for more startup options
  Please add the following to your Gemfile to avoid polling for changes:
    require 'rbconfig'
    if RbConfig::CONFIG['target_os'] =~ /(?i-mx:bsd|dragonfly)/
      gem 'rb-kqueue', '>= 0.2'
    end
```
And that's what I did. On BSD, the app does work with (no nagging message) and without this patch and I don't have a way to measure the performance impact. Yet, I understand what they mean, kqueue is the default kernel event signalling mechanism in BSDs and whenever I implement anything tcp related with kqueue.

manno:
Sadly this leads to a modified `Gemfile.lock` on Linux.

I tried an alternative approach in the bsd-kqueue branch, does this work for you?

erdgeist:
LGTM. Vielen Dank.

PR #343, opened 2017-09-09
Add EventClassifiers to conferences

erdgeist:
Each conference can define several dynamic categories called classifiers, along with some informative text, what this classifier is about.

Once at least one classifier is defined, all events in the conference can be assigned one or more dynamic category weighting, called EventClassifier. Users can select the appropriate Classifiers via checkboxes in the new/edit Event views and then use sliders to select the desired weighting between 0% and 100%.

EventClassifier values are displayed as meter and numerical indicators in the event view and are exported as a dictionary when rendered as json.


manno:
I just skimmed the code. Looks great and all tests are green. Two additional things, though:

* I think we don't need the files in  `test/fixtures` since we're using factory girl
* Do you mind to create a feature test for the UI parts? Something like this [cfp_can_edit_event_test.rb](https://github.com/frab/frab/blob/master/test/features/cfp_can_edit_event_test.rb). You might have to add `js: true` to the `test` definition, but probably not.

erdgeist:
Added UI tests with capybara. Test coverage for UI interaction is moderate but problems should immediately become apparent.

erdgeist:
Also fixed the tests for postgres, where some sequences didn't start with 1.

PR #350, opened 2017-10-01
Fix jbuilder based json responses

manno:
Probably due to updating to rails 5, the custom built
jbuilder JSON responses were no longer used.
This brings back jbuilder views and strenghtens testing around
JSON responses in the admin area.
What started as responses in different formats has now become an API
apparently.
XML responses continue to use ActiveModel::Serializer and might return
more or less fields that JSON.

### DIFF
--- a/app/controllers/cfp/people_controller.rb
+++ b/app/controllers/cfp/people_controller.rb
@@ -18,6 +18,8 @@ class Cfp::PeopleController < ApplicationController
     end
   end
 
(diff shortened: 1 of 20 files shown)


manno:
@erdgeist is anything missing from the events index? I added speaker_ids, event_classifiers and state (for orga only). The other fields, description, subtitle, should be there, too.

erdgeist:
LGTM, thanks.

Issue #358, opened 2017-10-15
Generate event tokens to allow other speakers to join an event


erdgeist:
Currently submitting an event as a team is tedious. Submitters add a "Submission Note" that verbally informs the orga about a second or third presenter. Often those presenters don't even have an account, yet.

My proposal is to generate ( one token per event | or one for each new speaker on demand ) to pass around that allows – when presented via an extra field in the "new event" view – to be added to the event. Probably it's easier to just pass an "add me to the event"-URL in CfP scope around; as this will only work with users that have logged in (i.e. that have an account) the login mechanism will ensure that only existing users can be added.

As usual, if this is appreciated, I would implement it.

manno:
I like it. 
One token per event should be suffcient. Maybe have add a 'join another event' button to the CFP interface. 

Using a direct URL could be done additionally, with the same token. I feel like people often have problems copying long URLs. The 'confirm talk' mail has an URL like that, which bypasses login. 

But either way, form or URL, I'm happy to have this feature. The current state is, as you said, very tedious.

Sadly we can't bypass login here, since we need to know who will be joining that talk.

erdgeist:
I think, everything I wanted to have is now in there in #359. After review this can be merged.

manno:
Merged in 0f3d646d2df94ba8bc56ced9cc33b3959278b9f6

PR #359, opened 2017-10-17
First shot at speaker invitation token

erdgeist:
This is a first shot at #358. Will do some translations and UI fixes.


erdgeist:
This now in a state where I'm a little clueless what exactly I need to test for. I assume that the has_secure_token logic just works. Would that be an integration test or a controller test for adding just works with tokens?

manno:
Since it's limited to one controller I think a controller test is a good start. If you want to test the whole procedure, probably integration is better suited. But whatever is easier is fine.

app/controllers/cfp/events_controller.rb
```diff
@@ -83,6 +83,17 @@ def confirm
     end
   end
 
+  def join
+    @token = params[:token] || ''
+    @event = @token.blank? ? nil : Event.find_by(invite_token: @token)
+    if request.post?
+      return redirect_to cfp_join_event_path, flash: { error: t('cfp.join_token_unknown', token: @token) } if @event.nil?
```
artemasmith: It's quite complex logic. I think it's better to split it
```ruby
def join
 return unless request.post?
 @token = params[:token] || ''
 @event = @token.blank? ? nil : Event.find_by(invite_token: @token)
redirect_on_invalid_event
redirect_on_already_created
@event.event_people << EventPerson.new(person: current_user.person, event_role: 'speaker')
redirect_to edit_cfp_event_path(@event), notice: t('cfp.join_success')
end

private

def redirect_on_invalid_event
    return redirect_to cfp_join_event_path, flash: { error: t('cfp.join_token_unknown', token: @token) } if @event.nil?
 end

def redirect_on_already_created
     return redirect_to edit_cfp_event_path(@event), notice: t('cfp.join_token_already_used') if @event.people.exists?(current_user.person.id)
end

Comment (erdgeist) [reply app/controllers/cfp/events_controller.rb]: The returns from redirect_on_invalid_event and redirect_on_already_created don't propagate up to the join function, so this does not work.

File: app/controllers/cfp/events_controller.rb [hunk 2]
```diff
@@ -83,6 +83,17 @@ def confirm
     end
   end
 
+  def join
+    @token = params[:token] || ''
+    @event = @token.blank? ? nil : Event.find_by(invite_token: @token)
+    if request.post?
```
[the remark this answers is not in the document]
erdgeist: Yep.

app/controllers/cfp/events_controller.rb
```diff
@@ -86,11 +86,16 @@ def confirm
...
     @token = params[:token] || ''
     @event = @token.blank? ? nil : Event.find_by(invite_token: @token)
-    if request.post?
-      return redirect_to cfp_join_event_path, flash: { error: t('cfp.join_token_unknown', token: @token) } if @event.nil?
-      return redirect_to edit_cfp_event_path(@event), notice: t('cfp.join_token_already_used') if @event.people.exists?(current_user.person.id)
-      @event.event_people << EventPerson.new(person: current_user.person, event_role: 'speaker')
-      return redirect_to edit_cfp_event_path(@event), notice: t('cfp.join_success')
+    return unless request.post?
+
+    if @event.nil?
```
artemasmith: Looks like more readable:)

erdgeist: Thank you. Important remark.

PR #360, opened 2017-10-18
Store original location to return to after signin

erdgeist:
Looks like devise has had a mechanism to return to the originally requested resource after sign in, as described here:

https://github.com/plataformatec/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update

So for root and the direct cfp sign in urls this patch sticks to the old logic of redirecting users to their edit profile page (why is that, btw) and for all other cases return to the original resource (as would come in handy with join-token uris).


manno:
I think the redirect to the 'edit profile' page is a leftover of the old CfP logic, trying to ensure speakers edit their profile before submitting a talk. Providing  a linear flow through the submission (register, profile, submit talk). We probably shouldn't redirect if there is already a profile from previous submissions.

But why did the 'return' break the tests?

erdgeist:
Looks like there's an integration test that checks for users being redirected to the cfp page after they sign up. I forgot the return in my final commit, so the return seems to break the test.

This means after I figured out where the user should finally be redirected to, I should update the test to reflect the change.

manno:
I think the cfp user should still be redirected to the profile page for the first time after sign up.

erdgeist:
Currently, they are also constantly nagged to enter their availabilities, even when the view doesn't exactly indicate, how.

I saw that you can redirect after sign-up, but what we actually need the direct after the confirm, right?

https://stackoverflow.com/questions/8003347/overriding-devise-after-sign-up-path-for-not-working

manno:
Yes, since we use `confirmable` in the users model. So the availabilities button is missing from the view?

erdgeist:
The button is there in the cfp/person view, but the flash is also shown when I then edit an event: e.g.
http://192.168.216.129:3000/en/brainiac/cfp/events/504/edit

erdgeist:
Found out, why the flash stays there: If you want to alert on the current view and not in the redirect, you need to set the flash.now variable, not the flash. So in app/controllers/cfp/people_controller.rb it must read

>       flash.now[:alert] = t('cfp.specify_availability')

Will look for more instances and then commit.

app/controllers/auth/sessions_controller.rb
```diff
@@ -19,6 +19,8 @@ class Auth::SessionsController < Devise::SessionsController
   protected
 
   def after_sign_in_path_for(resource)
+    goto = stored_location_for(resource)
+    return goto if not goto.blank?
```
artemasmith:
```ruby
return goto if goto.present?
```

artemasmith:
Excuse me, @erdgeist, why don't you squash commits for PR? IMHO, it's easier to find and revert changes, when 1 PR = 1 commit.

erdgeist:
@artemasmith Isn't squashing done when finally merging to master? I'm not familiar with best practises here but having work-in-progress commits visible until the feature is finished looks like a good idea.

artemasmith:
@erdgeist I'm sorry, I didn't mention that it's still in progress. Yep, you're right, commits are squashed before PR merge.

erdgeist:
Reviving the PR, as at the last conference there's been users complaining again about invite token URLs not working.

@eladeyal-intel can you please point me to where you managed to get the automated translations? I would want to provide them for the new flash message.

eladeyal-intel:
@erdgeist 

i18n-tasks translate-missing -f en -l de,es,fr,it,pt-BR,ru,zh

erdgeist:
@eladeyal-intel I see, this requires a paid account. Thanks.

eladeyal-intel:
@erdgeist - Maybe you can push English and Dutch; and after this is merged I will translate the rest with this tool.

Issue #356, opened 2017-10-10
Allow speakers to schedule their own talks


manno:
Open conferences might profit from speakers scheduling their own talks:

* add feature flag
* add scheduler to cfp view
* automatically accept submitted talks
* speaker can't edit talks like in the admin view
* speaker can't de-schedule other talks from other people
* publishing the schedule is still done by crew only, crew can still edit/decline events

fdel15:
Hey @manno is anyone working on this? I would like to tackle it if it's available.

manno:
He @fdel15 , nobody is working on that. Sorry for not having more detail on that, but I thought this could be like a 'self organized' feature toggle in the admin menu. If you toggle it, speakers get access to a schedule view (like the one in admin). Deactivate at any point to clean up and publish, or just leave it up to the participants.

manno:
@fdel15 did you get around working on this, otherwise I might start implementing this?

PR #369, opened 2017-11-10
Open conference

manno:
This implements issue  #356


saerdnaer:
@manno Can you notify me, when I can test this feature on frab.cccv.de or a test instance?

PR #376, opened 2017-11-23
Big translations update - Ready to review

mtancoigne:
As stated in #185, the current state of the translations lacks some elements.
I started to check every view one by one and replace the hardcoded string by translated ones.

We're working on this for our next conferences, so the changes are live on our server: https://evenements.agilemans.org

Here is the checklist:
- [x] - Translate views
- [x] - Override view from Devise and make them translatable
- [x] - Translate strings in controllers
- [x] - Translate strings in helpers
- [x] - Translate activerecord's fields names
- [x] - Automatic language selection from #379 

For now, we think that most of the translation work is done; if you're interested by this PR, tell me and I'll dispatch the new strings in existing files
- [ ] - Re-order strings in translations files

PR #379, opened 2017-11-29
Choose default locale based on user's browser settings

smortex:
Web browsers allow users to configure their preferred languages.  If Frab is translated in a language the users prefers to English, default to this language instead of English.

E.g. if my language settings are `['pt', 'en-US', 'en', 'en-GB', 'de']`, then `pt-BR` will be used (getting the settings properly is a bit tricky, hence the new gem dependency).

This only affects the choice when no language is set, and it's still possible to switch language as previously.


manno:
@ScottyTM I think frab was using the conferences locale instead of the browsers locale intentionally? Do you remember why/if this is useful behaviour? Or is this just another pentabarf relict?

Switching to the users locale seems more user-friendly :)

smortex:
Close this PR since it's commits where included into #376.

PR #380, opened 2017-12-06
Fix missing tz information on start_time

erdgeist:
Substitution did not take conference timezone into account. Quick fix to correct that.


jjeising:
Shouldn't that still use a locale based format?

PR #382, opened 2017-12-10
Translations - WIP

mtancoigne:
Please do not merge now;
This PR is the end of #376

It will provide a cleaner more complete translation template.

- [x] Merge every locales in their own file
- [x] Prepare subfolders
- [x] Sync every locale with each other to have a full base of work
- [ ] [In progress] Split translations in one external file per key (i.e. : `fr/devise.yml`, `fr/notifications.yml` and so on)


mtancoigne:
Weird, it seems the tests takes age to execute on Travis and locally :/

app/controllers/application_controller.rb
```diff
@@ -25,6 +25,8 @@ def page_param
   def set_locale
     supported_languages = %w[en de es pt-BR fr]
 
+    supported_languages.each { |l| I18n.load_path += Dir[Rails.root.join('config', 'locales', l, '*.yml').to_s] }
```
smortex: :warning: This is evaluated at each request.  I guess that `I18n.load_path` modification invalidates all cached data.  Maybe this could be moved into `config/initializers` which is only evaluated one at startup.

mtancoigne: Thanks for the tip, i'll take a look tonight.

config/application.rb
```diff
@@ -35,6 +35,9 @@ class Application < Rails::Application
     # Configure sensitive parameters which will be filtered from the log file.
     config.filter_parameters += [:password, :password_confirmation]
 
+    # Localization
+    I18n.load_path += Dir["config/locales/*/*.yml"]
```
smortex: `config/application.rb` is meant to configure [some determined settings](http://guides.rubyonrails.org/configuring.html#configuring-rails-components) **during** framework initialization, which might happen too early.

It would be safer to [wait for the framewrok to be loaded before altering it's configuration](http://guides.rubyonrails.org/configuring.html#using-initializer-files) in this case.

mtancoigne: You're certainly right, they use `config.i18n.load_path` in the [documentation](http://guides.rubyonrails.org/i18n.html). Even if the way I load the string is wrong, setting the path with `config.i18n.load_path` instead of `I18n.load_path` in application.rb should be good ?

smortex: Using `config.i18n.load_path` looks good, yes.  I only expect `config.foo = :bar`  directives to appear in `config/application.rb` :wink: .

test/integration/sign_up_test.rb
```diff
@@ -16,7 +16,7 @@ class SignUpTest < ActionDispatch::IntegrationTest
 
     post '/users', params: {
       'user' => { 'email' => 'test2@example.org', 'password' => 'frab12345', 'password_confirmation' => 'frab12345' },
-      'commit' => 'Sign up', 'conference_acronym' => @conference.acronym, 'locale' => 'en'
+      'commit' => 'Sign Up', 'conference_acronym' => @conference.acronym, 'locale' => 'en'
```
smortex: This is probably a translation lookup problem (sign_up is not found and is capitalized).  it worked before refactoring, so you should not change the tests ;-)

test/integration/sign_up_test.rb
```diff
@@ -31,7 +31,7 @@ class SignUpTest < ActionDispatch::IntegrationTest
     }
     follow_redirect!
     assert_includes @response.body, "#{@conference.title} - Call for Participation"
-    assert_includes @response.body, 'Update profile'
+    assert_includes @response.body, 'Update Profile'
```
smortex: This is probably a translation lookup problem (sign_up is not found and is capitalized).  it worked before refactoring, so you should not change the tests ;-)

neomilium:
@mtancoigne Please note, I made some improvements here:
https://github.com/frab/frab/pull/384

You probably will need to merge it into this PR.

test/integration/sign_up_test.rb
mtancoigne: The same key was used for signup, but tested "Sign Up" and "Sign up"...

mtancoigne:
I think we're good, all the strings are here.

The reference file is `config/locales/en/_en_all.yml`, and i'm thinking about merging some keys before splitting the big file in many smaller ones:
- `errors` key with `activerecords.errors`
- `activemodel` key with `activerecords` (but I have a doubt on this one)
- `emails` key with `emails_module`
- `mailers` key with `emails_module`
- `session` key with `navigation`

Can someone tell me if there's something wrong ?

neomilium:
IMHO, you should clean up this branch (squash or split some commits) to prevent from big squash during merge of PR.
Big commits (e.g. made due to PR squash) are really painful to read, understand and so on.
Additionally, it does remove timestamps and original commiters too which made `git blame` an inefficient tool.

neomilium:
@mtancoigne did you use tools to de-dup and/or merge yaml files ?

mtancoigne:
@neomilium, Yes, I have a plugin for atom that checks/sorts yml files, so I use it to verify the file structure. comparing translations is easy but takes time :).
About squashing, i'll do it when the PR will be ready, i'm only waiting for a reviews of the changes about merging keys. Ideally, in the end, there will be a file per key (i.e;.: events_module.yml, notifications.yml and so on). If we can optimize a bit before doing so, that would be great as there is currently one file per language.

neomilium:
@mtancoigne Do you know `i18n-tasks` gem ?
https://github.com/glebm/i18n-tasks

I give it a try on Frab and see thats quiet efficient.
This way, you only have to provide a config file according to your will then merging will be easier.

IMHO, automated ways are better than monkey patches.

neomilium:
Please note that we (@smortex and me) automated the translations files handling using `i18n-tasks`.
This tool is exactly what we need to improve i18n in Frab.
We provide this PR at #385

mtancoigne:
@neomilium Hi and sorry for the response time.
Well, if everything is fine,I close this PR :)

neomilium:
@mtancoigne I checked your commits and made a diff, between your fork with i18n-tasks applied and i18n-tasks branch based on upstream. There are a least one commit that improved things I did not.
It may be a good thing to help maintainers to apply PR #385 then offer a PR with your improvements.
I noted that i18n-tasks could be applied on your branch by cherry-pick 3135385744aa440a3570fe679fa37804633b183d then move config/locales/$language/_${language}_all.yml to config/locales/$language/$language.yml and finally run `bundle exec i18n-tasks normalize -p`

mtancoigne:
@neomilium I'll take a look in a day.

neomilium:
@mtancoigne I have some pending PR and some commits remains locally.
I'll let you a note when its done.

PR #388, opened 2017-12-23
Use translation keyword in cards.pdf

neomilium:
Allow translation of "Abstract" title


neomilium:
@erdgeist Any reason that prevents this fix merge ?

erdgeist:
Nope, have just been busy over the last days. Thanks!

PR #389, opened 2017-12-23
Preview track color in conference schedule

smortex:
The conference schedule page (not the public one) displays events using green, orange and red as background color depending the event having conflicts or other conditions to look after for organizers.

The track color is not visible to organizers from this screen, which was requested by our customer.

This Pull-Request add a small square that previews the event's track color as shown bellow:

Preview

### DIFF
--- a/app/assets/stylesheets/frab.css
+++ b/app/assets/stylesheets/frab.css
@@ -92,6 +92,14 @@ div.event {
   padding: 3px;
 }
 
+div.color-preview {
+  width: 1em;
+  height: 1em;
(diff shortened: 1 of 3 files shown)


erdgeist:
I think the square is a little clumsy. In the halfnarp companion app I use a 2px border two signal track, with a toggle button to change the whole background to track color. That would eat up less space inside the text areas. Should I port it to frab?

smortex:
@erdgeist I guess that would be awesome!

manno:
Sounds great, I merge this PR until we get the border?

PR #391, opened 2018-01-02
allow coordinators to preview the schedule

saerdnaer:
The have access to the schedule anyway. I would even go one step further and allow the reviewers (method crew_only) to access the preview, but this might require further discussion...


erdgeist:
LGTM. Thanks.

PR #397, opened 2018-01-09
Fix typo in "address" word

neomilium:
Please note that even `im_address` keywords were incorrect in EN and FR language files, there are correct in other languages and in source code.


erdgeist:
Thanks

PR #406, opened 2018-01-26
Use activerecord "human.attribute_name"

neomilium:
Affected keywords were missing. Instead of create duplicates I use already translated activerecord attributes.

### DIFF
--- a/app/views/conference_users/_table.html.haml
+++ b/app/views/conference_users/_table.html.haml
@@ -1,9 +1,9 @@
 %table.zebra-striped
   %thead
     %th.first
-    %th= t('activerecord.attributes.user.email')
-    %th= t('activerecord.attributes.conference.acronym')
-    %th= t('activerecord.attributes.conference_user.role')
+    %th= User.human_attribute_name('email')
+    %th= Conference.human_attribute_name('acronym')
+    %th= ConferenceUser.human_attribute_name('role')
     %th= t(:actions)
   %tbody
     - collection.group_by { |cu| cu.user_id }.each do |user_id, conference_users|
--- a/app/views/public/schedule/_events.html.haml
+++ b/app/views/public/schedule/_events.html.haml
@@ -7,8 +7,8 @@
       %thead
         %tr
           %th &nbsp;
-          %th= t '.title'
-          %th= t '.speakers'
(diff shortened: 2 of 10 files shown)


neomilium:
@smortex, @manno  I'm not sure if its a good practice to address this problem. Comments are welcome.

app/views/public/schedule/_events.html.haml
```diff
@@ -7,8 +7,8 @@
       %thead
         %tr
           %th &nbsp;
-          %th= t '.title'
-          %th= t '.speakers'
+          %th= t 'activerecord.attributes.event.title'
```
smortex: I would have used [human_attribute_name](http://api.rubyonrails.org/classes/ActiveModel/Translation.html#method-i-human_attribute_name):

```ruby
Event.human_attribute_name('title')`
```

smortex: That way, no need to be aware of how things are organized in i18n files to understand what the code is trying to do :stuck_out_tongue_winking_eye:

neomilium:
@smortex  Thanks! I updated my commit and update existing translation based on 'activerecord.attributes' using:

```
find app/views/ -exec sed -e "s/\(.*\)t.'activerecord\.attributes\.\([^.]*\)\.\(.*\)')*/\1\u\2.human_attribute_name('\3')/" -i '{}' \;
```

app/views/conference_users/_table.html.haml
```diff
@@ -1,9 +1,9 @@
 %table.zebra-striped
   %thead
     %th.first
-    %th= t('activerecord.attributes.user.email')
-    %th= t('activerecord.attributes.conference.acronym')
-    %th= t('activerecord.attributes.conference_user.role')
+    %th= User.human_attribute_name('email')
+    %th= Conference.human_attribute_name('acronym')
+    %th= Conference_user.human_attribute_name('role')
```
smortex: `Conference_user` is not a valid class name in respect to Ruby conventions.  Typo?

neomilium: Commit fixed. Thanks1

neomilium:
@manno This PR is now complete and can be merged.

PR #442, opened 2018-11-16
Anti-Invisibility of Non-male speakers in German ;)

laryllian:


manno:
This is only one of many places where the new German translation is awkward.
We had a short discussion and would prefer the generic female form.

Maybe use a neutral form for the profile pages: https://fahrplan.events.ccc.de/congress/2018/Fahrplan/speakers/8515.html

saerdnaer (approved):

PR #451, opened 2018-12-28
Remove HTML from event abstract

### LINKED ISSUE #390: Event Abstract: do not allow html tags
/cc @a-tze

manno:
TODO – fixed #390 

@saerdnaer 

- [ ] migrate existing data?


saerdnaer:
```
CREATE OR REPLACE FUNCTION strip_tags(TEXT) RETURNS TEXT AS $$
    SELECT regexp_replace($1, '<[^>]*>', '', 'g')
$$ LANGUAGE SQL;

UPDATE events SET abstract = strip_tags(abstract) WHERE abstract LIKE '%<%'; 
```

manno:
Hm, funktioniert das in PostgreSQL, MariaDB und sqlite?

saerdnaer:
@manno MySQL implementation of strip_tags(): https://stackoverflow.com/a/13347316/521791

manno:
`Event.where("abstract like '%>%'").each do |e| e.abstract = strip_tags(e.abstract); e.save; end`

manno:
Problem is some events have invalid syntax in their abstract:

* `<http://bla> some text <http://another url>`, gets deleted

We can merge this, but data migration can't be done easily in an automated fashion.

PR #453, opened 2018-12-30
Event video url

manno:
Drops the old video table and adds a single `video_url` field to lectures.

Rake task to import from podcast.xml (as provided by voctoweb) still exists, but writes to video_url instead of video and link list.


saerdnaer:
Probably easier to use that the old mechanism. We could think about using the embed code from voctoweb using oEmbed. But thats probably an own PR.

Issue #446, opened 2018-12-12
STARTTLS error when sending email


MisterGlass:
I have a new install up on AWS. I have it configured to use SES to send emails, but whenever the server attempts to send an email (such as during signup) it fails.

Logs:
```
INFO -- : Completed 500 Internal Server Error in 68ms (ActiveRecord: 9.6ms)
FATAL -- : 
FATAL -- :  Net::SMTPAuthenticationError (530 Must issue a STARTTLS command first):
FATAL -- : 
FATAL -- :  /usr/share/rvm/rubies/ruby-2.5.3/lib/ruby/2.5.0/net/smtp.rb:981:in `check_auth_response'
```

Config:
```
SECRET_KEY_BASE=REDACTED
RUBY_VERSION=2.5.3
FRAB_HOST=https://frab.example.org
FRAB_PROTOCOL=https
FROM_EMAIL=cfp-mailer@example.org
SMTP_ADDRESS=email-smtp.us-east-1.amazonaws.com
SMTP_PORT=587
SMTP_DOMAIN=frab.example.org
SMTP_ENABLE_STARTTLS_AUTO=true
SMTP_NOTLS=false
SMTP_USER_NAME=REDACTED
SMTP_PASSWORD=REDACTED
FRAB_CURRENCY_UNIT="€"
FRAB_CURRENCY_FORMAT="%n%u"
```

MisterGlass:
I dug into the config code a bit, found https://github.com/frab/frab/blob/master/config/initializers/action_mailer.rb#L12

Seems like the value of the env setting is ignored, and if this is at all present it effects things. That was not the behavior I expected.

I commented out the setting in my env and its now working.

MaxRink:
Im running into the same Issue but dropping SMTP_NOTLS from my env vars doesnt help

manno:
Should we change the intializer to?

```
 if ENV.fetch('SMTP_NOTLS', 'false') == 'true'
```
So it would be false if absent and only disable TLS, if set to 'true'?

PR #459, opened 2019-02-11
Remove SMTP_NOTLS ENV Var

MaxRink:
Remove the SMTP_TLS Enironment variable to restore the ability to use Mailservers with TLS enabled.
As described here ( https://github.com/frab/frab/issues/446#issuecomment-446471008 ) the existance of this Var disables TLS Support completely..
This change adds the ability to use TLS-Mail back into the Docker Image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,6 @@ ENV RACK_ENV=production \
     FROM_EMAIL=frab@localhost \
     SMTP_ADDRESS=172.17.0.1 \
     SMTP_PORT=25 \
-    SMTP_NOTLS=true \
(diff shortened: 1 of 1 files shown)


manno:
@MaxRink is it still possible to use the Docker image without SMTP TLS? Not sure if someone is using that?

MaxRink:
Well, the dockerfile sets the environment Variable SMTP_NOTLS to true.
I can Override this to false but looking at the code that interprets the ENV-Vars ( https://github.com/frab/frab/blob/master/config/initializers/action_mailer.rb#L12 ) the sole presence of this Var disables TLS-Support for good.
And as far as i know there is no easy way to unset an env var defined in an dockerfile so this needs to go to get it working again

Issue #159, opened 2015-03-12
Allow uploading of assets into an object storage


brianredbeard:
One of the few hurdles to having Frab be fully horizontally scalable is uploading of assets directly into an object store.  This is being opened to track the feature.

ScottyTM:
I'm not sure what you are trying to achieve. I can't really image a usecase where you would need frab to scale horizontally. frab is usually used by at the most double digit users at the same time. This load can be handled by a decent single machine. Also the other parts of frab are not really setup for a multi-machine install.

brianredbeard:
As this has progressed over time, the use cases have also evolved.  At the time this was originally opened, part of the concern was around practical experiences that CoreOS ran into using Frab for the CFP review for both CoreOS Fest and Tectonic Summit.

Beyond the horizontal scaling issue the use of object storage is of a very real practical usage when Frab is deployed in a Linux container.  This alleviates the need for any persistent filesystem from within the container.

To be clear, after our acquisition by Red Hat we are no longer using Frab, but I can provide more context (if needed) with the goal of increasing the usability for others.

manno:
Previous work for S3 with (now deprecated) paperclip: https://github.com/frab/frab/pull/160/commits/3053abd0f856102c89199a809b1fd9cd69cb2def

Can you be more specific on horizontal scaling issues (https://github.com/frab/frab/issues/464), what else is missing?

manno:
Directories which require persistent storage

* https://github.com/frab/frab/issues/418#issuecomment-440469347
* https://github.com/frab/frab/issues/466

WIP: https://github.com/frab/frab/pull/434

PR #460, opened 2019-02-18
frab does not scale horizontally

eladeyal-intel:
I think it's a good idea to mention this explicitly so that people don't try it.

My source is https://github.com/frab/frab/issues/159


eladeyal-intel:
@manno  ?

PR #498, opened 2019-08-19
Make max attachment size configurable

### LINKED ISSUE #493: Attachments file size is hard limited
Attachments file size is hard limited to 42MB.

https://github.com/frab/frab/blob/cbe681c60e8f5734acf10f116a8d8acefebd2498/app/models/event_attachment.rb#L6

@manno is there a reason to this specific limit? (i.e. something bad will happen if we try to store larger files)

Can we make this configurable?

eladeyal-intel:
Max attachment size is now configurable per conference. 

Closes #493


manno:
I'm sorry, I should have been more explicit. I would have opted to make this configurable *per server*, not per conference :)
Do you need it per conference, otherwise I'd add an env var.

eladeyal-intel:
Per server is good enough for me.

Issue #499, opened 2019-08-20
third party authentication


eladeyal-intel:
I created a bunch of changes that allow frab user authentication to be from 3rd party - LDAP server (closes #17) and Google (closes #20) for now. Others (github, facebook etc.) should be easy now.

I will submit them as different commits so it's easier to review the code. For starters I attach here the documentation for the final state - I think it should go to the wiki after merging.

:scissors::scissors::scissors:=== start documentation===:scissors::scissors::scissors:

Integration --> User authentication
=============

By default, frab let people to register to the frab webside by providing a username and a password. You can change that to support authentication via one or more third party services as well. It's important to notice that a user (identified by an e-mail address) can only be associated with one kind of a login. i.e., she can either login via frab's username/password form or via Google integration - not both. It is therefore best to select an authentication strategy before deploying to production.

Authenticate with Google
---------------------------
You'll need to get from Google a Client ID and a Client Secret. The process is something like this:

1. Go to 'https://console.developers.google.com'
2. Select your project.
3. Go to Credentials, then select the "OAuth consent screen" tab on top, and provide an 'EMAIL ADDRESS' and a 'PRODUCT NAME'

Then update the environment variables  `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`.

Authenticate with LDAP server
------------------------
LDAP integration is helpful when deploying frab in an enterprise, where you don't want new users to remember another password. To enable Sign in with an LDAP server, populate the following environment variables:
1. `NAME_FOR_LDAP=`  This name will appear in the frab login screen, following  "Sign in". This overrides the language-dependent value defined in the locale. This is usually the enterprise name or the name of the organization managing the access.
2. `LDAP_PROMPT_TITLE=` (optional) - This name will appear as the title in the LDAP credentials entry form. (i.e. `Enter your NASA username and password:`).
3. `LDAP_HOST=`
4. `LDAP_PORT=`
5. `LDAP_METHOD=` One of: `plain`, `ssl`, `tls`.
6. `LDAP_BASE_DN=`
7. `LDAP_UID=` or `LDAP_FILTER=` - `LDAP_UID` indicates which LDAP field is used to search users.
- Common values for enterprises are: `LDAP_UID=sAMAccountName` or `LDAP_UID=UserPrincipalName`
- Commonly `LDAP_UID=uid` for OpenLDAP
[...]

PR #502, opened 2019-08-20
Third party authentication

eladeyal-intel:
See #499


manno:
Hey this is great, a lot of people were asking for that.

I didn't have time to test it and I'm worrying about how easy it is to support this, e.g. when updating to Rails 6. It looks like this is no problem, what is your impression of the Gems, are they in widespread use and updated regulary?

eladeyal-intel:
I'm sorry I have no idea. This is more or less my first venture into Ruby & Rails. My work was mostly cut and paste from the how-to, most stuff went smoothly. If one specific strategy gem is not good, adding a different one is easy - take a look at the Google commit.

manno:
Omniauth Gems seem to be very popular (Sinatra, etc). 

I think https://github.com/cschiewek/devise_ldap_authenticatable has the advantage that we could also access the users ldap data. That would be nice to keep roles in a central place to share them with other systems.
However it also looks more complicated. I merging this now from the command line, because I commented `env.example` and rebased against master.

manno:
Merged in 1f219763

PR #530, opened 2019-09-03
configurable event types per conference

eladeyal-intel:
Conference settings now include two ways to express which event types
are allowed.

First we have the same of event_types we had so far; `lecture workshop podium lightning_talk meeting film concert djset performance other`. These names are localized when used. Each of these can be allowed/disallowed individually per conference. The default is that they are all allowed - same as before.

Also, admin can add any number of additional types by specifying them in a semicolon delimited list. This list is not supposed to be localized.

I decided to store the allowed types as a string and not separate tables in the database in order to be less intrusive. For example admin can remove a type from the allowed types list; but any event with that type will keep it.

image

image

image

image

see frab#302


erdgeist:
This looks like something that I would love to see, but the current PR is a mess. There's changes in 67 files, most of them having nothing to do with "configurable event types".

If you manage to cherry pick those changes to a new PR, I will review and merge them before they start departing from master again.

eladeyal-intel:
@erdgeist - done.

erdgeist:
Thank you.

PR #555, opened 2019-09-24
add assistant

eladeyal-intel:
A new user role in an event: "assistant" . Assistant has the same
permissions as a speaker, but their availability does not count.
The assistant is notified by mail of changes to the event, similar to
presenter.

Prospective assistants can join events by use of the join token,
similar to co-speakers.

see frab#488


app/controllers/cfp/events_controller.rb
```diff
@@ -100,23 +100,26 @@ def confirm
 
   def join
     @token = params[:token] || ''
+    @join_as = params[:join_as] || ''
```
erdgeist: Wait a minute. Here you just copy the POST parameter join_as from the request … right?

app/controllers/cfp/events_controller.rb
```diff
@@ -100,23 +100,26 @@ def confirm
...
-
     if @event.nil?
         return redirect_to cfp_join_event_path, flash: { error: t('cfp.join_token_unknown', token: @token) }
     end
 
     if @event.people.exists?(@person.id)
       redirect_to edit_cfp_event_path(@event), notice: t('cfp.join_token_already_used')
     else
-      @event.event_people << EventPerson.new(person: @person, event_role: 'speaker')
+      @event.event_people << EventPerson.new(person: @person, event_role: @join_as)
```
erdgeist: … and here you just pass the event role from what you got from the request above? What stops you making yourself coordinator or submitter?

This doesn't look right.

eladeyal-intel: @erdgeist you are right of course. Thanks for spending time on these reviews..

I googled it and limiting parameters to specific value sets does not seem to be something that "strong parameters" deal with. So I ended up adding [this](https://github.com/eladeyal-intel/frab/blob/4c882f10047bb978db0fe35319fbf066653da471/app/controllers/cfp/events_controller.rb#L108):

```ruby
raise Pundit::NotAuthorizedError unless @join_as.in? ['assistant', 'speaker']
```

Basically anyone with a frab login can join to any event as a `submitter` or an `assistant` (and now: nothing else) ; all she needs is the invitation token. `submitter` can joins only during the CfP; `assistants` can join later too.

app/controllers/cfp/events_controller.rb
```diff
@@ -100,23 +100,28 @@ def confirm
 
   def join
     @token = params[:token] || ''
+    @join_as = params[:join_as] || ''
     @event = @token.blank? ? nil : Event.find_by(invite_token: @token)
 
+    return unless request.post?
+    
+    raise Pundit::NotAuthorizedError unless @join_as.in? ['assistant', 'speaker']
```
erdgeist: Wouldn't that rather be JOINABLES?

eladeyal-intel: fixed to

```ruby
raise Pundit::NotAuthorizedError unless @join_as&.to_sym.in? EventPerson::JOINABLES
```

erdgeist:
Basically, the PR looks good to me, but I would want @manno 's opinion as well. Also, I strongly dislike the term "stakeholder" for multiple reasons. Maybe ON_SITE or ON_VENUE would be a more neutral descriptor for that group?

eladeyal-intel:
Thanks erdgeist.

Re ON SITE: they're not necessarily on site. I see the assistant role used in two ways: (1) the speaker has a secretary which needs to get all relevant emails and is in charge of the speaker's travel arrangements and so forth. (2) a mentor capacity, someone who helps the speakers make sure their slides are attractive, well written, etc. She needs access to all files, and needs to know if the event was moved/rescheduled etc.

I think the term "stakeholder" is almost never used externally (user facing). I think I can change it to "people" or something.

erdgeist:
I wonder, if the term "subscriber" properly transports the meaning. If you look at why we build this list in the first place, is to ensure everyone who needs to be notified about state changes of the event … actually is notified. This is what a subscription is. So I think, this term might be used instead of steakholder ;)

eladeyal-intel:
I thought "watchers" to be confusing is the sense that people come to a conference in order to watch the actual event.

So I changed all "stakeholders" to "subscribers" - internally the term was not used.

I kept the external wording "send to 2 people" and did not use "2 subscribers" becuase I think it's clearer.

PR #556, opened 2019-09-25
fix GroupingError with psql

eladeyal-intel:
closes frab#554


eladeyal-intel:
I tested this with psql and with sqlite3. I can't test it with mysql.

Issue #518, opened 2019-08-29
submission grading with multiple metrics


eladeyal-intel:
I implemented "review metrics". Reviewers can provide a grade to each submission based on each metric; and we can see average grading just like the average overall metric. The idea is that it allows reviwers to communicate more values than a single star rating; so the selection process can be done in a more fine-grained way.

I'm not big on "review metrics" as a term, but I'm getting into it. Couldn't find anything better though.

A side effect of this GUI is that it needs to be possible to save an incomplete review form. So a reviewer can write the text comment and do some of the ratings but not all of them. An event is considered "reviewed" if and only if there's a rating in the overall rating prompt (stars system). The review metrics are optional to fill in.

### Settings for conference now invites the admin to specify the review metrics 

The default is : No review metrics.

![image](https://user-images.githubusercontent.com/20379005/67045284-a3ca8f80-f136-11e9-9f47-2be04a52fa70.png)
 

![image](https://user-images.githubusercontent.com/20379005/67045370-d4aac480-f136-11e9-8aca-cbede1ca1aab.png)

### Rating

When providing rating, reviewer can also provide a grade between 1 to 5 to each review metric, or leave as N/A. It is now possible to provide only a comment and/or only a review metric, without assigining a star rating.

Coordinators/admin can see the list of all reviews for the event like they can see all ratings.

![image](https://user-images.githubusercontent.com/20379005/66721929-8e8cf280-ee10-11e9-91c9-3f76bd03af70.png)

The "Events --> Event Ratings" screen now show, for each event, the average of each review metric. The value of N/A is not taken as a part of average. Sorting events by a review metric ascending/descending is supported.

The "Event Rating Count" counts only the number of people who  provided a value for the star rating.

![image](https://user-images.githubusercontent.com/20379005/66722015-7a95c080-ee11-11e9-82af-4e59a5e1fdd1.png)

eladeyal-intel:
![image](https://user-images.githubusercontent.com/20379005/64076107-007a0400-ccc9-11e9-8823-3a7f7e000ddb.png)

eladeyal-intel:
Modifications to the manual:

✂️ ✂️ ✂️
 Conference Crew / Admin => Organizer => Setting Up a New Conference => add section "Review Metrics"

[...]

PR #557, opened 2019-09-26
Submission Review (grading) with multiple metrics

eladeyal-intel:
see frab#518


eladeyal-intel:
Added test for editing an event score's and checking that the average is calculated correctly.

erdgeist:
There's a lot going on in this PR that doesn't have to do with review metrics, like introduction of the assistant role and the concept of stakeholders. This makes it quite hard to review the review metrics part.

Since I am not sure how this is supposed to work, anyway, I focussed on anything BUT the review metrics and this looks okay.

eladeyal-intel:
@erdgeist - I updated #518 to show better what "review metrics" are supposed to do.

eladeyal-intel:
- Changes the settings tab to be called "Reviewing" and not "Review Metrics". Updated #518 accordingly.
- Simplified ReviewMetric::safe_name so now it has a default value for names only included non-latin characters.

app/models/event.rb
```diff
@@ -3,7 +3,7 @@ class Event < ApplicationRecord
   include ActionView::Helpers::TextHelper
   include EventState
   include HasEventConflicts
-
+
```
erdgeist: whut space?

app/views/event_ratings/show.html.haml
```diff
@@ -75,6 +75,17 @@
         %fieldset.inputs
           = f.input :rating, as: :rating
           = f.input :comment, input_html: {rows: 3}
+          = f.simple_fields_for :review_scores do |ff|
+            = ff.input :score,
+              as: :radio_buttons,
+              collection: [[t('events_module.not_applicable'), '0'], ['1', '1'], ['2', '2'], ['3', '3'], ['4', '4'], ['5', '5']],
```
erdgeist: Does the zero value mean that you can undo a rating?

eladeyal-intel: Yes. The value 0 is displayed as N/A when rating; and as a "blank" when looking at the rating summary. It does not taken account when calculating average review.

image

erdgeist:
I understand most of the PR now and see little changes that have the potential not be corrected easily if they contain mistakes I have not spotted.

I wonder how good the interface will actually work in the event reviews pane, but these may also be patched up later.

So from me that would be a go. Maybe @manno wants to comment, if not I'll merge after 24h.

eladeyal-intel:
https://github.com/frab/frab/pull/614 is this pull + a modification to add_fake_data to create fake reviews, if you want to experiment with the GUI

Issue #523, opened 2019-08-31
attachment overview


eladeyal-intel:
![image](https://user-images.githubusercontent.com/20379005/66413259-b7237f80-e9ff-11e9-842e-8b3faacd2efd.png)

I'm working on "attachment overview" screen, where crew can get a look on all the attachments uploaded to the conference. Attachments with pre-set titles (see #522) appear in their own columns and are indicated by the upload date. The idea is for admins to be able to track which events already uploaded (or updated) the files they are supposed to and which submitters need to be reminded to complete their material and upload it.

Feedback appreciated

eladeyal-intel:
The manual was updated: https://github.com/frab/frab/wiki/Manual/_compare/bb0bd0bfd868b7ba65075efc030cb064c87c1820...efac6fc44f4d7f44429dad19e1d64cb626731b86

PR #561, opened 2019-10-02
add "Attachment overview" screen

eladeyal-intel:
With this view crew can get a look on all the attachments uploaded to
the conference. Attachments with pre-set titles appear in their own
columns and are indicated by the updated date.

see #523


eladeyal-intel:
changed the tab caption to "Attachments" instead of the longer "Attachment overview".

Added i18n

added test

Issue #566, opened 2019-10-03
harmonize mail templates and notifications


eladeyal-intel:
1. Mail templates will use the same dynamic variables as "Conference Notifications". A migration will automatically upgrade existing templates to the new format.

2. If admin use only fields which are related to the person and not to the event, then the person will get a single e-mail, even if she participates in multiple events. If admin uses fields which are related to specific events, then people will get one e-mail per event.

3. Mail templates can have dynamic subject as well.

![image](https://user-images.githubusercontent.com/20379005/66107408-e0d84300-e5c8-11e9-9fb1-5dbcb14e9119.png)

eladeyal-intel:
Merged in #578. Closing.

PR #568, opened 2019-10-03
More flexible mail templates

eladeyal-intel:
see #566


eladeyal-intel:
need to fix send_bulk_mail_job.rb - on production I get `ArgumentError wrong number of arguments (given 0, expected 2)`

PR #569, opened 2019-10-03
some automatic translations

eladeyal-intel:
created with this magic command

i18n-tasks translate-missing -f en -l de,es,fr,it,pt-BR,ru,zh


erdgeist:
LGTM 👍

Issue #577, opened 2019-10-06
Add count (and sub-count) to index views


erdgeist:
It would be cool to have a quick overview about the number of Events and Persons in their respective Lists, something along the lines of

<h1>Listing 999 events</h1>

and for Filtered Views

<h1>Listing 123/999 events</h1>

After looking into the code, I don't find an obvious place to get both values to the template. Any idea?

eladeyal-intel:
@erdgeist See #570 I started this. I think you can see my code there. Does this make sense ?

erdgeist:
That's good. Can you please create a PR out of this commit? I can't find a way to do that from the interface.

eladeyal-intel:
@erdgeist I think it's time to close this ticket.

eladeyal-intel:
Still fixed. Still needs to be closed.

PR #579, opened 2019-10-08
Improved event filters

eladeyal-intel:
This PR adds a few improvements to filterings.

1. Add filtering for numeric values  i.e., average rating, number of ratings, average review score.

2. The tables and the "filters" pane are now subject to i18n

image

3. For text filters - capability to filter several options together; i.e., show all events in "rejected" and "rejecting" state.

image

4. Harmonization of the filter logic so the same code is re-used.

5. Format change for the active filter list.  Clicking on the filter opens the modal filter editor (for both text and numeric filters)

image

image

6. Filter icon in the table headers - gray if not used (not being filtered by this coloumn) ; blue outline if being filters with this coloumn. It changes color to full blue when hovering the mouse, as an invitation to click.

I rewrote everything in this PR and edited it in steps so it should be easier to review.

### DIFF
--- a/app/assets/images/filter_highlight.svg
+++ b/app/assets/images/filter_highlight.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
+<svg width="28.347pt" height="28.347pt" fill-rule="evenodd" overflow="visible" stroke-linejoin="bevel" stroke-width=".501" version="1.1" viewBox="0 0 28.347 28.347" xmlns="http://www.w3.org/2000/svg">
+<g transform="scale(1 -1)" fill="none" stroke="#000">
+<g transform="translate(0 -28.347)">
+<g stroke="#0003ff" stroke-linejoin="miter" stroke-miterlimit="79.84">
+<path d="m10.695 14.102-9.03 12.537h24.198l-9.119-12.64v-7.063l-6.049-4.537z" fill="none" stroke-width="1.512"/>
+<path d="m10.695 14.102v-11.703l6.049 4.537v7.063l9.119 12.64h-24.198l9.03-12.537z" fill="#0003ff" stroke-linecap="round" stroke-width=".5"/>
+</g>
+</g>
+</g>
+</svg>
--- a/app/assets/images/filter_off.svg
+++ b/app/assets/images/filter_off.svg
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
+<svg width="28.347pt" height="28.347pt" fill-rule="evenodd" overflow="visible" stroke-linejoin="bevel" stroke-width=".501" version="1.1" viewBox="0 0 28.347 28.347" xmlns="http://www.w3.org/2000/svg">
+<g transform="scale(1 -1)" fill="none" stroke="#000">
+<g transform="translate(0 -28.347)">
+<path d="m10.695 14.102-9.03 12.537h24.198l-9.119-12.64v-7.063l-6.049-4.537z" fill="none" stroke="#b2b2b2" stroke-linejoin="miter" stroke-miterlimit="79.84" stroke-width="1.512"/>
+</g>
+</g>
+</svg>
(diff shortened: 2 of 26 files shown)


erdgeist:
It would be great to rebase all of the PRs currently pending on master. Currently most of them seem to belong in a certain order and I can't merge the newer ones for lack of experience with what the earlier commits do.

eladeyal-intel:
@erdgeist   This PR is based on #518 ; becasue his PR adds filters for numeric values in the event list; and #518  adds numeric values in the event list (=review averages). If needed I can change the merge order.

eladeyal-intel:
As part of this commit, I also harmonized the different filters' display properties. See https://github.com/frab/frab/blob/53537f123d5d06dee72d52aa1aaef0275062b030/app/helpers/events_helper.rb#L28-L58 

Today I changed some of the function names and changes the filter data to return a ruby `Struct` (previously was an array) to be clearer.

erdgeist:
Okay, time to rebase ;)

erdgeist:
Looking through the patch, I see several UX sins already. My [x] hack has been bad enough and I've always wanted to replace it by some more intuitive way like the ❌ sign or other graphical element easy enough to spot. But hiding the comparison operator behind a toggle character is just way too unintuitive. I'd suggest (and later implement once this PR is aligned with master) to visually group each filter with a decent border and some darker fill, and make the operator a drop down:

X [ value ≥ 2 ]
   [ value > 2 ]
   [ value ≤ 2 ]

so that you immediately see that there are some options and what will be the option after you select it.

eladeyal-intel:
@erdgeist 

How about this? The X icon removes the filter; the ⭮ icon cycles between  ≥ = ≤ ; the link "multiple" opens the multi-value selector

image

erdgeist:
Why would only "multiple" open the selector? Allowing the same on "Confirmed" in that case would add consistency to the interface. Also, why the ⭮ icon, if you can actually show what happens next (i.e. the operator that would be selected if you click on the right button)?

But besides that, I like where this is going. Really looks a lot tidier than my initial hack.

erdgeist:
Even better: If there's only three operators, you can just put the two NOT currently selected in the box to the right so that you do not have hidden options.

What I still find irritating is that you can't change the numeric value we compare against. You have to manually look for an entry with a matching value to start this filter. From an interface perspective: Why would you use the "average of 3.4" as a starting point to compare against? Find the submissions that are equally(?) well ranked? Or find all submissions that are ranked better than the one shown here?

I come to think that since you obviously now have a good model of modals, clicking on these numbers in the list should open a modal allowing to select the operator and narrowing down the exact number to filter against. But maybe I mis-understood the workflow you had in mind.

eladeyal-intel:
We can have this:

image

and I can probably add the third operator ≥ as a second button next to the = so that there are no hidden options.

The behaviour I was thinking about is that the moderators will sort the submissions based on one criteria, then filter out everything worse, then continue. I'm having a bit of a trouble imagining a modal window that would make sense

erdgeist:
The more I think of it, the less sense makes adding a filter by clicking on the number. The way I see it would be to add a filter symbol next to the title of the column and then add a modal that behaves differently for counts (like the event rating count) and review averages.

I can see why you would want to select based on "worse than this average", e.g. so you can bulk-update them to rejected or send bulk mail. But I have a hard time imagining why you would want to filter "everything with more than two reviews" or "exactly two reviews", when you actually only would want to sort to see the ones that have received the least attention, so that they pop up at the top of the list.

erdgeist:
While I am at it: I'm maintaining a companion tool for a while that does exactly what you're working on here and I've received quite some feedback on it. It gets the events per track from frab in a cron job, and then renders a list that can be filtered with a type-ahead filter, sorted by all review categories plus their combined average, sorted by the amount of reviews, sorted by the own review count (0 or 1) and sorted randomly – so that every reviewer can work through a different set of events when starting from the top).

The tool is here https://erdgeist.org/gitweb/c3rater/tree/ and if you drop me a DM, I can show you a link to a live demo version.

eladeyal-intel:
OK I almost got it:

image

Can you DM  in GitHub? I didn't know. Can you e-mail me by adding changing my username - replace - to @ and add `.com`

erdgeist:
I just noticed, there's even a public documentation for the tool :) https://content.events.ccc.de/rater.html

eladeyal-intel:
@erdgeist I updated this using modal filter window for both text and numeric.

erdgeist:
This looks and works amazing. Just two more pixels below the filters bar and I'd merge it. :)

erdgeist:
Just a little question, the "Showing XX of YY Events" we were talking about in #577 went away?

eladeyal-intel:
I added a few pixels below the filter bar. is this what you mean?

image

https://github.com/frab/frab/blob/21b431c05684bcc67eb7092552ad5f863c743243/app/assets/stylesheets/frab.css#L253

erdgeist:
I just noticed that when your session expires and you klick the filter icons, nothing happens (because the modal route is not answered?) until I reloaded the page and then was asked to log in again.

erdgeist:
Looks like the last merge broke your CSS, a closing bracket is missing.

eladeyal-intel:
> I just noticed that when your session expires and you klick the filter icons, nothing happens (because the modal route is not answered?) until I reloaded the page and then was asked to log in again.



I added a page refresh in case of modal load error; so in the case @erdgeist described, user will be presented with login screen instead of no response.

https://github.com/frab/frab/blob/0be612dd34f4d4e595b19a8632f00bcad36f9134/app/views/events/_filters_modal_holder.html.haml#L11-L12

> Looks like the last merge broke your CSS, a closing bracket is missing.

fixed

erdgeist:
Amazing work. Thanks.

eladeyal-intel:
Thanks :-)

Issue #581, opened 2019-10-09
Batch editing for events


eladeyal-intel:
The description was moved to #582

eladeyal-intel:
merged

PR #582, opened 2019-10-09
Bulk edit

eladeyal-intel:
batch editing for events - see #581


app/views/events/_below_table.html.haml
```diff
@@ -13,4 +13,88 @@
...
+                       class: 'danger'
+          = simple_form_for(:bulk_edit_add_person, 
+                            url: batch_actions_events_path(request.query_parameters), 
+                            method: :post, 
+                            data: { persons: Person.fullname_options } ) do |f|
+            .peoplefilter
+              = f.label t('events_module.bulk_edit.add_person')
+              = text_field_tag :filter, 'filter', style: "width:200px"
+              = select_tag :person_id, 
+                           options_for_select(Person.all.sort_by(&:full_name).map{|p| [p.full_name_annotated, p.id]}),
```
erdgeist: This is a really bad idea. I usually hate opening the "Edit persons" view on an event, because it takes ages to load and render, due to the thousands of users in frab. They need to be queried, put in the selector and the type ahead usually is rather slow as well. I've always wanted to make this into some sort of AJAX thingy querying the server on demand after three characters have been typed.

Having this huge list delivered on every event index, wether you want to bulk edit it or not, is simply unacceptably wasteful. It may work on test databases, but after managing more than one or two conferences in the same instance, this is annoying.

app/views/events/_below_table.html.haml
```diff
@@ -13,4 +13,88 @@
                        id: 'bulk_email',
                        data: { confirm: t('notifications_module.send_unspecified_notification_confirm', count: @num_of_matching_events) }, 
                        class: 'danger'
+     
+    - if policy(@conference).manage?
+      %p
+        = link_to t('events_module.bulk_edit.toggle'), "#bulk_edit", data: { function: 'toggle', args: { target: '#bulk_edit' } }
```
erdgeist: The whole interface looks a little rough. I wonder why each dropdown needs it's own Set button, when you can easily just make the whole thing a form and then submit all changes at once. It can even show what to edit in the dropdown as first placeholder text:

[ Bulk set state ] [ Bulk set track ] [ Bulk set type ] [ GO ]

app/views/events/_below_table.html.haml
erdgeist: I see where this is coming from. You want to set an assistance to all the matching lectures. I wonder if that can be better achieved by a modal overlay, also leaving more space.

I will then look into how to make the type ahead work better. Honestly, the current implementation is a kludge. You never see while typing ahead if you actually narrowed the list down already.

eladeyal-intel:
Updated the interface to be modal. 

This should fix the problem of slowing down the event list screen. 

We can fix the person selection & filter to be faster later.

image

image

image

image

image

eladeyal-intel:
I think this is ready for merge now. Maybe @erdgeist can translate 'Edit %{count} event:' inside de.yml; apparently google translate can't think on something better.

erdgeist:
I still find the model dialogue visually displeasing. Would you accept a commit trying to tidy up the interface?

eladeyal-intel:
Sure

eladeyal-intel:
added bulk editing for event duration. (i.e., change all events from 45min to 40min)

image

eladeyal-intel:
If and when   #670 is merged I will change this to use Ajax person filter too.

eladeyal-intel:
Hi @erdgeist. If the ui still displeased you, you can provide a mock up for how you think it should look and I'll try to implement that.

eladeyal-intel:
I can accept that this is not the best UI for the mission and I'm open for feedback and implementing change. Barring that I think the capability to edit multiple events instead of one by one individually  is important and helpful and I would want to see it merged.

eladeyal-intel:
Hi @erdgeist, @manno - would you like to review this? TIA

saerdnaer (approved):
LGTM

PR #578, opened 2019-10-06
More flexible mail templates

eladeyal-intel:
see frab#566

PR #583, opened 2019-10-09
Filter then email

eladeyal-intel:
allows user to filter events (in the all events / my events / ratings screen) and then send a mail template to people involved with those events.

Merge notes: 
- This must be based on #578 (because the change in email sending)


erdgeist:
Okay, next PR. Can you please resolve the merge conflicts, so that I can review this one.

eladeyal-intel:
@erdgeist Can we work on #578 first? #578 changed the way mass mails are sent.

eladeyal-intel:
rebased!

app/models/event_person.rb
```diff
@@ -114,6 +114,10 @@ def default_url_options
   def to_s
     "#{model_name.human}: #{person.full_name} (#{event_role})"
   end
+  
+  def self.for_events(events)
```
erdgeist: Where is that used?

test/integration/view_event_test.rb
```diff
@@ -51,8 +54,9 @@ class ViewEventTest < ActionDispatch::IntegrationTest
   end
 
   test 'finds events for search term' do
-    get "/#{@conference.acronym}/events?q%5Bs%5D=track_name+asc&term=frap&utf8=%E2%9C%93"
+    get "/#{@conference.acronym}/events?q%5Bs%5D=track_name+asc&term=#{@conference.events.last.title.split.last}&utf8=%E2%9C%93"
     assert_response :success
     assert_includes @response.body, 'frap'
```
erdgeist: This assert is not guaranteed anymore, once you don't use the search term 'frap' anymore. Or am I missing something?

app/controllers/events_controller.rb
```diff
@@ -132,6 +144,33 @@ def start_review
...
+    redirect_back(alert: t('ability.denied'), fallback_location: root_path) and return if mail_template.blank?
+    
+    events = search @conference.events_with_review_averages.includes(:track)
+    event_people = EventPerson.where(event_id: events.to_a.pluck(:id))
+    
+    if Rails.env.production?
+      SendBulkMailJob.new.async.perform(mail_template, event_people)
+      redirect_back(notice: t('emails_module.notice_mails_queued'), fallback_location: root_path)
+    else
+      SendBulkMailJob.new.perform(mail_template, event_people)
```
erdgeist: Why would development use the more robust approach of synchronous delivery? I would hate seeing Bulk Mail working in dev mode and then wonder why this doesn't work anymore with frab users in the wild …

Isn't that something that should be configurable independently of production/development settings?

app/views/events/index.html.haml
```diff
@@ -1,15 +1,17 @@
 %section
   .page-header
     .pull-right
-      - if policy(@conference).manage?
+      - if policy(@conference).manage? and !showing_my_events
         = action_button "primary", t('events_module.add_event'), new_event_path, hint: t('events_module.add_event_hint')
```
erdgeist: Why wouldn't you allow adding new events in the "my" view?

app/views/events/index.html.haml
```diff
@@ -1,15 +1,17 @@
...
         = action_button "", t('events_module.print_cards'), cards_events_path(format: :pdf), hint: t('events_module.print_cards_hint')
         = action_button "", t('events_module.print_accepted_cards') , cards_events_path(accepted: true, format: :pdf), hint: t('events_module.print_accepted_cards_hint')
-    %h1= t('titles.event_list')
+
+    %h1=t('titles.event_list') unless showing_my_events
+    %h1=t('events_module.my_events') if showing_my_events
 
   %ul.tabs
```
erdgeist: I wonder if we can also turn those into includes as with the _events_tabs.html.haml … I am asking, because ratings.html.haml does not have the events_path(events: :my) but my_events_path

test/integration/view_event_test.rb
eladeyal-intel: The events created by this test are `Introducing frap part x`  - see https://github.com/frab/frab/blob/master/test/factories/event.rb#L2

So I filter by the last word (the serial number) and we'll see the title `Introducing frap part x` for one event anyway.

app/controllers/events_controller.rb
eladeyal-intel: I copied from https://github.com/frab/frab/commit/8e2404c7eb9c11063995ea6c4657f467fded87fb#diff-2f9d0e8ee584f04039c864ea0da162daR14-R22 . It was there for 5 years.

I think the reason they do synchronously for deelopment is that it will let it open the shim e-mail as a new browser window.

app/views/events/index.html.haml
eladeyal-intel: I was keeping legacy behavior. The "My Events" page didn't have this button. 

I will remove the condition. i.e., I will add the "Add Event" button to "My Events" page.

app/models/event_person.rb
eladeyal-intel: never used. I will remove it.

app/views/events/index.html.haml
eladeyal-intel: I pushed a new version which combines all the tabs to a separate file. https://github.com/frab/frab/blob/5cb9f31ea56230550f5ff2417537f570862fc2db/app/views/events/_event_list_tabs.html.haml

app/controllers/events_controller.rb
erdgeist: Ah, I see, thanks. So I hope someone did think this through in the past. ;)

runtest.sh
```diff
@@ -0,0 +1,9 @@
+RAILS_ENV=test rails test test/features/filtering_event_list.rb test/features/editing_event_review_test.rb test/integration/view_event_test.rb
```
erdgeist: This is new. How does that work? Is this your local test code committed by mistake?

eladeyal-intel: Mistake

Issue #17, opened 2012-05-31
Need LDAP authentication support


synnack:
Hi, we're currently looking to use frab for OHM (Hack-Tic event in 2013 in the Netherlands in the line of HIP, HAL, WTH and HAR)..

But we need to integrate the authentication with LDAP, as the rest of our infrastructure does. Is this possible?

hackman:
Hmm we still need LDAP authentication support for SHA2017 the next incarnation of OHM2013.

BurtGummer:
is there LDAP authentication support in frab now?

manno:
No, there isn't. I'm currently working on an upgrade to rails 5 and will bring devise back. Devise can be extended to use LDAP I think.

zonque:
Using devise is certainly a good idea, and most of the data model should still fit I guess. I wanted to do the transition some time ago, but then realized that frab was originally based on devised, but then ditched it. Could you explain what was the reason for that?

oneiros:
> is there LDAP authentication support in frab now?

As I mentioned above, authenticating against existing users in an LDAP directory is quite trivial. But this is only part of the solution. There is also the question, if all users in LDAP should be able to log into frab and what role they should assume. Of course you could also create user accounts in frab first and only use LDAP as an alternative authentication mechanism for these users.

But the biggest hurdle imho is new users. frab allows for new users to sign up. So in an LDAP scenario you would need either one of two things:

1.) frab would need additional code to create new user records in LDAP. This is _not_ trivial, especially since every LDAP directory is different. Consequently the LDAP integrations for devise I have seen do not offer this.

2.) As an alternative, frab could provide an option to disable new signups.  But this would of course render some of frab's best features useless.

> I wanted to do the transition some time ago, but then realized that frab was originally based on devised, but then ditched it. Could you explain what was the reason for that?

This is ancient history, but for the record:

frab has a use-case that devise did not support very well back in the day: frab has different interfaces that are seperated from one another, the admin interface for organizers and the call for papers interface for submitters. Submitters should of course only see the latter, but organizers sometimes need to submit proposals, so they need be able to access both interfaces.

I do not recall the specific problems I had, but I do remember very well, that I had to wrestle with devise on more than one occassion and still failed to make it work the way I wanted it to. I spent hours in the debugger, stepping through code trying to understand the flow. devise is based on warden, a rack middleware, and simply trying to find out what way a username and password take through the code during login made my head hurt.

[...]

PR #596, opened 2019-10-12
Test for LDAP integration

eladeyal-intel:
This PR adds an automatic test for LDAP integration (ref #17 and https://github.com/frab/frab/pull/502).

The test is directly against the LDAP server from [forumsys](https://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/) - I think the CI volume will not be too harsh on that site.

On the other hand, if and when that server goes offline, frab CI will break.

So I'm not sure if it makes sense to merge this or not.

An alternative way to test LDAP in CI is [here](https://cweiske.de/tagebuch/ldap-server-travis.htm); but it seem to carry a bunch of complications with it. i.e., the test would be different for CI than what you would run on your dev station; and there's an elaborated bring-up sequence.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,13 @@
 FRAB_HOST=frab.test
 FRAB_PROTOCOL=http
 FROM_EMAIL=frab@frab.test
+NAME_FOR_LDAP=free testing server at ldap.forumsys.com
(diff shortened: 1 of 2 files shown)


eladeyal-intel:
moved to #663

Issue #454, opened 2018-12-30
Prevent certain modifications on events after soft deadline


erdgeist:
Currently, submitters can change event duration even after the CfP has finished, this can lead to frustration in putting together a schedule and conflicts introduced by interested parties during the conference.

Also using the Join-Event link should only be valid until the soft deadline has passed, currently we only check for the hard deadline which means that even after a lecture has been accepted, more speakers could be added.

manno:
Sounds great, I would put that behind a feature toggle as not every conference has a big team of coordinators.

So if `conference.freeze_accepted` (or something) is set:

- [ ] speaker cannot update event duration
- [ ] join event links are disabled if hard deadline is hit

erdgeist:
After discussion with 36c3 content teams we also understand that changing tracks and event type should also not change after either hard deadline has occurred or event has been set to an accepting state.

eladeyal-intel:
Am I right to assume all these limitations would apply to app/views/cfp/events/_form.html.haml only ? So this is the form used by submitters. Orginizers and coordinators use app/views/events and they should be able to change these fields indefinitely.

PR #603, opened 2019-10-21
add joinlink option to outgoing mail template

eladeyal-intel:
This PR adds the ability to add `%{joinlink}` to outgoing mail template, which will attach a URL to join the event as a speaker (or after #555 as assistant). So the accept email can include somthing like this: 

```
Please ask your co-presenters to join the event by clickling this link: %{joinlink}
```

Additionally this PR changes the default behavior if the link cannot be generated (for `%{link}` and `%{joinlink}`). The existing code just leaves `%{link}`. The new code will substitute this for a hyphen, for example:

```
Please ask your co-presenters to join the event by clickling this link: -
```

In order for the links to work correctly for both production & dev, this PR adds a new Environment variable `FRAB_PORT`. Together with `FRAB_PROTOCOL` (which was not used) and `FRAB_HOSTNAME`, they are used for URL generation.

# ---
image

```
Sent mail to roberts.brett@example.com (310.2ms)
Date: Mon, 21 Oct 2019 19:18:05 +0300
From: frab@localhost
Reply-To: shermanshanahan@hermancremin.com
To: roberts.brett@example.com
Message-ID: <5dadda3d9457d_43a3faab0216250808a@BIG.mail>
Subject: Your submission to Spectre IX Conference
Mime-Version: 1.0
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: 7bit
locale: en
title: Spectre IX Conference

Please ask your co-presenters to join the event by clicking this link: http://localhost:3000/en/spectre-ix/cfp/events/join/bWfgtzCA12LC193UhAXUSYww
```

image


erdgeist:
Even though this looks good to me, in my use case I wouldn't want to allow speakers to join after the event has been accepted. The amount and identities of speakers are what content curation teams base their acceptance on. So the join link shouldn't even work after CfP has ended which by it's nature is before bulk acceptance emails are sent ;)

But this, of course, might be handled differently at other conferences so I am not against it. However, once you're obviously expecting it to work that way, you might want to check if joining after CfP ended actually works for you.

This might also be a per-conference option:

Allow join:
- Never
- Until soft deadline
- Until hard deadline
- Until accepted

I kind of remember seeing something about this in issue #454.

eladeyal-intel:
@erdgeist  this PR is just about adding a way to send the request join link by e-mail. The code for allowing/rejecting a join is not affected. The link proposed here does not bypass these tests, it will still bring prospecting users to the "join this event" page.

Currently the join_as only allowed before hard deadline:

https://github.com/frab/frab/blob/46376763e6a0f39fc9f8b733ea21d80e3e2f315a/app/controllers/cfp/events_controller.rb#L107

https://github.com/frab/frab/blob/46376763e6a0f39fc9f8b733ea21d80e3e2f315a/test/features/event_join_test.rb#L39

BTW- workaround - one can disable all the join links in a conference by running this in the rails console:

    Conference.find_by(acronym: 'ripcord').events.update_all(invite_token: nil)

Issue #575, opened 2019-10-05
Import profile image from omniauth


eladeyal-intel:
When authenticating with Google, frab can take the profile image from the google account. Google assigns a default image to users without a profile image, so this is what will be used.

![image](https://user-images.githubusercontent.com/20379005/66260674-6b3bc500-e7ca-11e9-8a99-9ead1e08c184.png)

If anyone has an LDAP server in the wild with images, let me know. I took whatever our enterprise LDAP server delivers.

The environment variable `OVERRIDE_PROFILE_PHOTO` is used to configure priorities. If set, then the Google (or LDAP) image overrides frab image on every login. If the environment variable is not set, then image is copied if there's no avatar defined for the user.

eladeyal-intel:
https://github.com/frab/frab/wiki/Manual/_compare/0fa4cad3e1ddd337ce48a355cba94aa2df6b466d...87da22033c18dae92c3ac564d859f5f0c4ff727d

PR #606, opened 2019-10-23
Import profile image from omniauth

eladeyal-intel:
see #575


eladeyal-intel:
For #693 - image URL may be in: `request.env['omniauth.auth'].extra.raw_info.picture`

PR #607, opened 2019-10-23
normalize i18n files

eladeyal-intel:


eladeyal-intel:
Normalization was achieved using `i18n-tasks normalize`.

Issue #490, opened 2019-05-26
Handle events submitted after soft deadline


towo:
1. There should be some kind of indication that an event was submitted after the soft deadline for an event.
2. The review mode should also have a toggle for only showing in-deadline events.

towo:
I might just implement this myself, would be good practice.

manno:
You could add the `created_at` date to the event view

eladeyal-intel:
@towo you can add `created_at` and add a sentence `(after soft dead-line)` in orange and even `(after the submission dead-line)` in red.

eladeyal-intel:
@towo merged in #626. I propose to close this ticket.

towo:
Agreed, @eladeyal-intel.

PR #626, opened 2019-11-02
highlight late submissions

eladeyal-intel:
1. Add ORANGE tag if event was submitted after cfp end; or RED tag if submitted after hard deadline. The tag appears in event lists in a short form, and in an explicit form in the event details page.

2. Change creation date display to conference timezone.

see #490

image

image

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -1010,6 +1010,13 @@ de:
     events_in_other_conferences: Events in anderen Konferenzen
     feedback_not_enabled: Das Feedback-System ist nicht aktiviert.
     hide_all: Alles verbergen
+    indicator:
+      full:
+        submitted_after_hard_deadline: Eingereicht nach harter Frist
+        submitted_after_soft_deadline: Eingereicht nach Fristablauf
+      short:
+        submitted_after_hard_deadline: Spät
+        submitted_after_soft_deadline: SPÄT
     inputs:
       hints:
         do_not_record: Soll dieses Event aufgezeichnet werden?
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -1032,6 +1032,13 @@ fr:
     events_in_other_conferences: Évènements dans d’autres conférences
     feedback_not_enabled: Le système de retours n’est pas activé.
     hide_all: Tout cacher
+    indicator:
+      full:
+        submitted_after_hard_deadline: Soumis après la date limite
+        submitted_after_soft_deadline: Soumis après la date limite
+      short:
+        submitted_after_hard_deadline: En retard
+        submitted_after_soft_deadline: EN RETARD
     inputs:
       hints:
         do_not_record: Est-ce que l’évènement sera enregistré ?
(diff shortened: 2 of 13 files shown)


config/locales/fr.yml
```diff
@@ -1032,6 +1032,13 @@ fr:
     events_in_other_conferences: Évènements dans d’autres conférences
     feedback_not_enabled: Le système de retours n’est pas activé.
     hide_all: Tout cacher
+    indicator:
+      full:
+        submitted_after_hard_deadline: Soumis après la date limite
+        submitted_after_soft_deadline: Soumis après la date limite
+      short:
+        submitted_after_hard_deadline: En retard
+        submitted_after_soft_deadline: EN RETARD
```
erdgeist: These do not look right.

config/locales/pt-BR.yml
```diff
@@ -1019,6 +1019,13 @@ pt-BR:
     events_in_other_conferences: Eventos em outras conferências
     feedback_not_enabled: O sistema de feedback não está ativado.
     hide_all: esconda tudo
+    indicator:
+      full:
+        submitted_after_hard_deadline: Enviado após prazo final
+        submitted_after_soft_deadline: Enviado após o prazo
+      short:
+        submitted_after_hard_deadline: Atrasado
+        submitted_after_soft_deadline: ATRASADO
```
erdgeist: These neither ;)

config/locales/de.yml
```diff
@@ -1010,6 +1010,13 @@ de:
     events_in_other_conferences: Events in anderen Konferenzen
     feedback_not_enabled: Das Feedback-System ist nicht aktiviert.
     hide_all: Alles verbergen
+    indicator:
+      full:
+        submitted_after_hard_deadline: Eingereicht nach harter Frist
+        submitted_after_soft_deadline: Eingereicht nach Fristablauf
```
erdgeist: from config/locales/de.yml, in German that would be "end_date: Enddatum
        hard_deadline: Harte Einreichungsfrist"

And I guess, these translations break conventions in other languages, as well.

eladeyal-intel: Your feedback is not clear enough for me. The "indicator full" is the string going for the event page, and the "indicator short" is the text going to the list pages. I thought just writing "Late" is enough becasue I did not want it to be too verbose

erdgeist: Sure, the short version is okay. But the long version of the term "hard deadline" and "soft deadline" already has had other translations. And now these mismatch. I showed you with the German example. Other translations don't even show differences between the two translations.

eladeyal-intel:
I moved the tag to the "created" column; so now it shows only on `All Events` and `My events` but not on `attachmetns` tab or `event ratings` tab

image

image

image

config/locales/de.yml
eladeyal-intel: So its

submitted_after_soft_deadline: Eingereicht nach Enddatum
submitted_after_hard_deadline: Eingereicht nach Harte Einreichungsfrist

?

PR #628, opened 2019-11-03
List PaperTrail versions per event

erdgeist:
This PR adds a "history" tab to the events view, listing all versions of the event and associated events.

It also cleans up the tabs on event views, so I didn't have to manually add the history tab to each of the others.

A review would be quite appreciated.


eladeyal-intel:
Some comments after reading the code (didn't try it yet)

- Did you forget to include `history.html.haml` ?
- I think the history view should exclude the review and feedback models (i.e. EventRating, EverentReviewAverage, ... ) for two reasons. (1) I think they will spam the list (2) it's a way to circumvent pundit authorization to see this data.
- I propose to move the tab order (from left to right) to be: Event details, people, history, Rating, Feedback
- Related bug you might want to fix while at it: Adding/removing/changing classifier for an event is not stored in papertrail.

erdgeist:
- Added the template.
- As the history is meant for coordinator/organizer only, anyway, I think it should be limited to those users in the first place. Leaving the spamming issue. I don't consider ratings spam. This is exactly the kind of changes I'd expect in this place. So basically we only need to exclude feedback. That could be handled by an extra :verbose parameter.
- Changed the tab order and, you're right, looks more natural
- Will look into adding PaperTrail to classifiers in another PR

eladeyal-intel:
@erdgeist I would think you'd want reviewers to see the history of the events they're reviewing, but not other events. I don't think it's important to track old review scores (on the contrary maybe). Anyway I'm not sure the reviews are integrated to paper-trail completely anyway.

erdgeist:
I see no easy way to exclude Ratings and Feedback. Updates types are saved as yaml key/value pairs deep in the PaperTrail version object with no efficient way to filter for them.

That being said, I actually didn't find it distracting seeing these verbose updates in a live conference, as they do happen at very specific times in an event's life cycle, making it actually pleasing to follow it.

PR #635, opened 2019-11-06
Silence devise warnings in tests

erdgeist:
I got a lot of troubling warnings:

>  DEPRECATION WARNING: [Devise] `DeviseHelper.devise_error_messages!`
> is deprecated and it will be removed in the next major version.
> To customize the errors styles please run `rails g devise:views` and modify the
> `devise/shared/error_messages` partial.
>  (called from block in _app_views_devise_registrations_new_html_haml__403213825357176102_17391016020 at /usr/home/erdgeist/frab/app/views/devise/registrations/new.html.haml:13)

and was told to `rails g devise:views`. Which I did. I then noticed there's a mixture of erb and haml templates lingering around, so I used the guide here https://github.com/plataformatec/devise/wiki/How-To:-Create-Haml-and-Slim-Views to consolidate all templates to haml.

Then I removed the html.haml templates for the mailer, to avoid multipart mails being sent.

One test called `sign_up_steps` had to be fixed, because it couldn't find the password field anymore. But now all tests work fine and without devise warnings.

Next up: Possibly removing unneeded templates.

### DIFF
--- a/app/views/devise/confirmations/new.html.haml
+++ b/app/views/devise/confirmations/new.html.haml
@@ -1,21 +1,13 @@
-- content_for :title do
-  = link_to t('page_titles.resend_confirmation'), root_url, class: 'brand'
-
-%section
-  .page-header
-    %h1
-      = t('titles.resend_confirmation_instructions')
-  .row
-    .span16
-
-      = simple_form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: { method: :post }, defaults:{error: false}) do |f|
-
-        = devise_error_messages!
-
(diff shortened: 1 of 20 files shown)


erdgeist:
Okay, it looks like the converter did not catch all cases of i18n strings, so it's back to the drawing board.

eladeyal-intel:
See #705

eladeyal-intel:
#705 was merged and it covered most of this PR. @erdgeist I think you can close this PR.

eladeyal-intel:
closing due to inactivity

PR #645, opened 2019-11-13
Add substitution for eventid

erdgeist:
For last Congress, we prepared speaker-specific check lists that would link to their respective frab event pages. Those checklists were rendered per event.

But when sending out acceptance mails, there was no way to automatically link to the document, since the event-id could not be inserted.

This PR simply adds the event_id variable to the list of substitutable parameters.

### DIFF
--- a/app/models/event_person.rb
+++ b/app/models/event_person.rb
@@ -79,6 +79,7 @@ class EventPerson < ApplicationRecord
     locale = person.locale_for_mailing(event.conference)
     string = s.gsub '%{conference}', conference.title
     string.gsub! '%{event}', event.title
+    string.gsub! '%{event_id}', event.id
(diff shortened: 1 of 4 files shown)


eladeyal-intel:
Wouldn't it be better to provide direct link as a substition I.e. %{eventpagelink} ?

erdgeist:
> Wouldn't it be better to provide direct link as a substition I.e. %{eventpagelink} ?

No. Imagine the link being https://conference.name/external-tool-checklist-%{event_id}.html. For example https://content.events.ccc.de/cfp/36c3/checklist-10816.html. I have rendered a handy overview page for each accepted event.

eladeyal-intel:
👍

erdgeist:
@eladeyal-intel Would you please do me the favour and add missing translations?

Issue #642, opened 2019-11-10
cannot restore backup without conference_review_metrics to freshly installed version with #557 


pcfe:
@eladeyal-intel
I have a backup that I created without *conference_review_metrics* (taken at the time of #598 ), when I try to restore that on a current frab version that is empty (so one where #557 and #549 are in), then I get

```bash
$ oc rsh frab-104-l9tr9
$ bash
I have no name!@frab-104-l9tr9:~$ . .env.production 
I have no name!@frab-104-l9tr9:~$ RAILS_ENV=production rake frab:conference_import --trace
** Invoke frab:conference_import (first_time)
** Invoke environment (first_time)
** Execute environment
** Execute frab:conference_import
[ ] restore conference
    conference: [REDACTED]
[ ] restore all conference_tracks
[ ] restore conference_cfp
[ ] restore conference_ticket_server
[ ] restore all conference_rooms
[ ] restore all conference_days
[ ] restore all conference_languages
[ ] restore all conference_review_metrics
rake aborted!
Errno::ENOENT: No such file or directory - bs_fetch:open_current_file:open
/usr/local/bundle/gems/bootsnap-1.4.1/lib/bootsnap/compile_cache/yaml.rb:49:in `fetch'
/usr/local/bundle/gems/bootsnap-1.4.1/lib/bootsnap/compile_cache/yaml.rb:49:in `block in install!'
/opt/frab/app/app/lib/import_export_helper.rb:321:in `restore_multiple'
/opt/frab/app/app/lib/import_export_helper.rb:191:in `restore_conference_data'
/opt/frab/app/app/lib/import_export_helper.rb:88:in `restore_all_data'
/opt/frab/app/app/lib/import_export_helper.rb:69:in `block in run_import'
/usr/local/bundle/gems/activerecord-5.2.2.1/lib/active_record/connection_adapters/abstract/database_statements.rb:259:in `block in transaction'
/usr/local/bundle/gems/activerecord-5.2.2.1/lib/active_record/connection_adapters/abstract/transaction.rb:239:in `block in within_new_transaction'
/usr/local/lib/ruby/2.6.0/monitor.rb:235:in `mon_synchronize'
/usr/local/bundle/gems/activerecord-5.2.2.1/lib/active_record/connection_adapters/abstract/transaction.rb:236:in `within_new_transaction'
/usr/local/bundle/gems/activerecord-5.2.2.1/lib/active_record/connection_adapters/abstract/database_statements.rb:259:in `transaction'
[...]

PR #648, opened 2019-11-15
Fix 642

eladeyal-intel:
Addresses #642 

1. Fix exception in the welcome page, if there's a conference with `start_date` or `end_date` = null.

2. Fix silenced exception if conference `start_date` and `end_date` calculation

3. Add an integration test which created a conference, exports it, imports it to a new name, and compares the new and old copies. This test fails without the other fixes in this PR.

4. `frab:conference_import` worked by disabling many callbacks, but never enabled them back. This caused tests to fail (because necessary callbacks were disabled). Now the import will re-eanble the callbacks it disables.


erdgeist:
Thank you!

Issue #608, opened 2019-10-25
Limit set of allowed lecture durations in CfP


erdgeist:
We're receiving a lot of submissions with impossible lecture durations, just because the dropdown allows it and people feel like they need to properly fit the length.

It would be cool to have a way of limiting what the user is presented with in the original CfP interface, even though the edit event view should offer to change duration later to the whole list of durations.

eladeyal-intel:
How about this:
 
![image](https://user-images.githubusercontent.com/20379005/67569755-64a0cd80-f738-11e9-8128-7e15e47380fb.png)

erdgeist:
Well, we actually need to provide some options, because e.g. at 36C3 we offer 40 and 60 minute lecture lengths, as well as longer slots for games and DJ sets and entertainment. I'd just love to de-select the 10, 20, 30, 50, 70, 80, 100 … as options so that the submitter can't pick wrong ones when initially creating their lectures.

eladeyal-intel:
How about this.

![image](https://user-images.githubusercontent.com/20379005/67572977-f8c26300-f73f-11e9-95a0-3ffd950f0161.png)

When migrating, we will auto-populate this field with the default, based on the max duration and the time slot configuration. But for new conferences, this will default to blank. I'm taking the assumption here that nobody wants an arbitrary long duration list, and that conference crew will be content with setting the list themselves.

eladeyal-intel:
actually I can use javascript to dynamically add & remove checkboxes for optional durations.

erdgeist:
I also thought that … because conferences have time slots and there's a maximal timeslot configuration, we could just add a set of checkboxes that are selected by default, representing timeslots that are shown in the "Create Event" interface. This should be 5 rows of 10 checkboxes for 10 minute slots and 500 minutes max duration. I wonder how acceptable this is UX-wise.

eladeyal-intel:
I have the conference settings ready.

The "default duration" is always checked and disabled (un-clickable)

![image](https://user-images.githubusercontent.com/20379005/68995840-e0ef8200-089a-11ea-848a-9409355b8dce.png)

eladeyal-intel:
duplicate of #304

eladeyal-intel:
#651 is merged - this issue should be closed.

Issue #304, opened 2017-03-18
It should be possible to configure fixed timeslots


cko:
At our conference we wanted to use frab to submit lightning talks (5min), talks (30 or 60 min) or workshops (2 or 4 hours). Having configured a timeslot duration of 5 Minutes results in submission with a wide range of durations. It would be great if it would be possible to configure fixed time slots instead of a time range for events, so that submitter can't choose from 00:05, 00:10, ..., 04:00 but just from a few fixed time slots like 00:05, 00:30, 00:60, 02:00, 04:00.

manno:
Here is how to replace the current mechanism with custom timeslots:

```
aee2189c - custom time slots (45/90/180) - Mario Manno (4 months ago)
diff --git a/app/helpers/events_helper.rb b/app/helpers/events_helper.rb
index cb01ddfb..b525af2f 100644
--- a/app/helpers/events_helper.rb
+++ b/app/helpers/events_helper.rb
@@ -24,6 +24,14 @@ module EventsHelper
     slots
   end
 
+  def custom_timeslots
+    slots = Array.new
+    [3, 6, 12].each do |i|
+      slots << [format_time_slots(i), i]
+    end
+    slots
+  end
+
   def format_time_slots(number_of_time_slots)
     duration_in_minutes = number_of_time_slots * @conference.timeslot_duration
     duration_to_time(duration_in_minutes)
diff --git a/app/views/cfp/events/_form.html.haml b/app/views/cfp/events/_form.html.haml
index bd597248..479229a6 100644
--- a/app/views/cfp/events/_form.html.haml
+++ b/app/views/cfp/events/_form.html.haml
@@ -9,7 +9,7 @@
           %b=@event.speakers.map(&:public_name).join(", ").html_safe
     = f.input :event_type, collection: translated_options(Event::TYPES)
     = f.association :track, collection: @conference.tracks, hint: 'Die Veranstaltung soll im Programm folgender Rubrik zugeordnet werden.'
-    = f.input :time_slots, as: :select, collection: timeslots, hint: t("cfp.event_time_slots_hint")
+    = f.input :time_slots, as: :select, collection: custom_timeslots, hint: t("cfp.event_time_slots_hint")
     = f.input :number_of_repeats, as: :select, collection: 1..4, hint: '1-4'
     = f.input :language, as: :language_select, only: @conference.language_codes, include_blank: true
     = image_input_box(f.object.logo)
[...]

PR #651, opened 2019-11-16
Limit set of allowed event durations in CfP

eladeyal-intel:
This PR address #608 and #304 :

1. Add a configurable set of event duration acceptable in cfp. The list of options is computed based on the "timeslot duration" and the "max number of timeslots". The "default event duration" is always acceptable. For conferences created before applying this PR, the default is that all duration are acceptable (same as legacy behavior). For new conferences, the default is that only the "default event duration" is allowed, and other durations need to be opt-in.

2. cfp event form is changed to only accept these values. Crew's event form is not affected.

3. cfp event form will not ask users to select duration if there is only 1 option available.

4. Rename "timeslots" to "duration" since we always present the value in minutes, never in timeslots.

5. Similar to "timeslot_duration" and "default event duration" : "allowed event durations" are configurable via the parent conference in case of a sub-conference.

6. Add validator, that a conference "default timeslot" (default duration) must be <= "max timeslots".

image

image

image


emirisman:
This would make lots of stuff easier. Is there more work required or could this be merged?

eladeyal-intel:
Ready!

PR #662, opened 2019-11-21
CI test for migrations

eladeyal-intel:
This PR:

1. adds to the CI `db:rollback` and `db:migrate` for the last few migrations. So this validates that migrations can apply and can revert on all supported databases.

2. On `sqlite` only, it verifies the resulting schema is the same as the original. This validates that new migrations are compatible with the updated `db/schema.rb`.

3. It removes a usage of unicode characters in one test, because it failed on `mysql` in some scenario. The same fix is also in #675 .


eladeyal-intel:
@manno do you know why the CI won't run for this PR? It wouldn't even start.

Issue #659, opened 2019-11-20
omniauth 1.8.1 gem not updateable to 1.9.0


manno:
The `omniauth-ldap` gem is depending on the old version. I don't think we're affected, but Github lists `omniauth` 1.8.1 as vulnerable to CSRF: https://github.com/advisories/GHSA-ww4x-rwq6-qpgf

@eladeyal-intel any ideas?

eladeyal-intel:
This link has a chapter "Mitigating in Rails applications" . I already implemented 1 and 2 when incorporating omniauth.

eladeyal-intel:
Except that I have 

```
protect_from_forgery :except => (...) :ldap
```



I don't remember why I needed that. I'll look into it

eladeyal-intel:
I'll try to replace the gem to https://rubygems.org/gems/gitlab_omniauth-ldap (3M downloads, last updated feb 2019)

manno:
Thanks for looking into this :)

eladeyal-intel:
OK I can now now explain why I think `protect_from_forgery :except => (...) :ldap` is acceptable.

The way `omniauth-ldap` works, is that it will present a form for the user to key in the username and password. The submit address for this form is the callback url
https://github.com/omniauth/omniauth-ldap/blob/3242f85968f26938d9a6ce028e511e788cc54722/lib/omniauth/strategies/ldap.rb#L30

So when `Users::OmniauthCallbacksController::ldap` is called
https://github.com/frab/frab/blob/ec0ad794b1b097e404b68bc6f19e6f096b7de578/app/controllers/users/omniauth_callbacks_controller.rb#L17 , it has the parameters `username` and `password` (in plaintext).

So in order to be admitted, the attacker would need to provide the correct username and password and can log the intended user in. So there's no attack.

eladeyal-intel:
..or we can wait for https://github.com/omniauth/omniauth-ldap/pull/86

eladeyal-intel:
@manno ?

eladeyal-intel:
#663 was merged - I propose to close this issue.

PR #663, opened 2019-11-21
replace omniauth-ldap with gitlab_omniauth-ldap 

eladeyal-intel:
see #659 

This PR:

(1) Adds a bunch of tests for LDAP login, using the free testing server at ldap.forumsys.com.

(2) replaces `omniauth-ldap` gem with `gitlab_omniauth-ldap` gem, which allows upgrading upgrading `omniauth` from 1.8.1 to 1.9.0 

(3) Adds support for `LDAP_DISABLE_VERIFY_CERT`ification; Apparently `omniauth-ldap` did not verify server certificate by default, and `gitlab_omniauth-ldap` does verify it by default.


eladeyal-intel:
I had to add `LDAP_DISABLE_VERIFY_CERT=1` because it wouldn't work with my enterprise LDAP server. Now it works. I think it's ready for merging.

PR #665, opened 2019-11-22
add fake profile images

eladeyal-intel:
from https://randomuser.me

image

image


erdgeist:
I don't think it's a good idea to randomly add links to external services to tests. Can't we just provide some demo images locally and link to them?

eladeyal-intel:
@erdgeist thanks for taking the time to review this.

It's not a "random service" - it's a service which is explictly set up for this pupose. The faces FAQ is explicitly permissive to use these faces for "Mock ups", which is what the data is used for. 

I looked at what "Faker" is doing - they also link to external services, they're not keeping images in their repository.

Note that the images are copied upon running `rails frab:add_fake_data`, so after that the images are safe within the test db.

I added a begin-rescue block so in the event this site becomes unavailable/down, the fake users are created sans photo.

eladeyal-intel:
@erdgeist i can try https://github.com/guinslym/Uifaces instead. Do you think it's better?

ScottyTM:
> @erdgeist i can try https://github.com/guinslym/Uifaces instead. Do you think it's better?

http://uifaces.com seems to no longer exist.

eladeyal-intel:
plus! It does not really affect tests, since the `add_fake_data` is never used as part of the test suite.

PR #670, opened 2019-11-23
AJAX-based person filter

eladeyal-intel:
Following a comment from @erdgeist I tried to rewrite the person-filter so that it is server-based and thus more responsive in deployments with large number of people in the database.

Up to 99 names are sent to the web client at a time, in which case they are listed in the dropbox selection box:

image

image

If there are too many matches, a message is displayed encouraging the user to keep typing

image

Once the text typed is specific enough to indicate an individual, the dropbox is replaced with a text box. There's no need to press ENTER to accept it.

image

Typing a person ID in the filter box works too - but it assumes the ID is complete. (So typing 5 will not match id 51 or 52)

 
image


eladeyal-intel:
Added test

saerdnaer (approved):
LGTM

eladeyal-intel:
Thank you @saerdnaer . I will merge this in 7 days if I don't get any feedback to the contrary.

PR #673, opened 2019-11-24
add duration as template content

### LINKED ISSUE #45: announce event length in accept message
If there's a fuckup or reorganization, the acceptance mail should state how long the slot for the lecture/event/whatever is, so that the speaker has the info.

eladeyal-intel:
fixes #45

image


eladeyal-intel:
@erdgeist ?

erdgeist:
Yep sorry, needed to take a break from content planning and content planning software after 36C3. Looking into the PRs now.

PR #681, opened 2019-12-19
fix #483

### LINKED ISSUE #483: "events that are unconfirmed" reports lists is partial
Reports -> "events that are unconfirmed" (in code: `unconfirmed_events`) lists *lectures* which are *unconfirmed*. 

See https://github.com/eladeyal-intel/frab/blob/master/app/controllers/reports_controller.rb#L40-L41

It should be renamed "unconfirmed lectures". (in code: `unconfirmed_lectures`)

eladeyal-intel:


erdgeist:
I am not sure about this: As long as we do have workshops, movies, etc, the term "lecture" is incorrect. Maybe "submission" would be more fitting?

eladeyal-intel:
@erdgeist take a look at https://github.com/frab/frab/blob/3a6fd32ffedb226dd93c00f4cae2235c824996f7/app/controllers/reports_controller.rb#L41 . The report in question specifically counts "lectures". But it's mislabeled. I was working on the assumption that someone needed this report.

PR #687, opened 2020-01-24
hotfix: Unbreak Dockerfile: Explicitly set ruby to 2.6.5

eladeyal-intel:
This PR:

(1) Changes `Dockerfile` to use `ruby:2.6.5` instead of `ruby:latest` (which is today 2.7 and does not work)

(2) Adds ruby 2.6.5 to the CI to help finding out if the docker version breaks.

See #688 
It's probably better to merge #690 instead of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: trusty
 rvm:
   - 2.4.4
   - 2.6.3
+  - 2.6.5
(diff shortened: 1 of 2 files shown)


eladeyal-intel:
#690 was merged so this is no longer needed. closing.

PR #700, opened 2020-02-17
fix frab#683

eladeyal-intel:


eladeyal-intel:
I'm not sure if this is the best way to fix this but at least it removes the exception.

PR #704, opened 2020-03-02
remove deprecation message from papertrail

eladeyal-intel:
See https://github.com/paper-trail-gem/paper_trail/pull/1176/commits


eladeyal-intel:
Apparantely this fix just replaces the depracation mesage with another one:

DEPRECATION WARNING: At some point early in the rails boot process, you have set PaperTrail.enabled = false. PT's rails engine is now overriding your setting, and setting it to true. We're not sure why, but this is how PT has worked since 5.0, when the config.paper_trail.enabled option was introduced. This is now deprecated. In the future, PT will not override your setting. See https://github.com/paper-trail-gem/paper_trail/pull/1176 for discussion. (called from <main> at config/environment.rb:5)

PR #711, opened 2020-03-07
WIP: upgrade sass-rails to v6, sprockets to v4, sass to sassc

eladeyal-intel:
I'm not sure I did this correctly. It seems to work on platforms I can test with. I'm not sure if `vendor/assets/fonts` needs to be specified in `app/assets/config/manifest.js` and how.


kotzw:
stale. Please close.

PR #724, opened 2020-09-15
Longer allowed_event_timeslots_csv db field

manno:
Fixes 'Mysql2::Error: Data too long for column'

### DIFF
--- a/db/migrate/20191116151919_add_allowed_event_timeslots_to_conferences.rb
+++ b/db/migrate/20191116151919_add_allowed_event_timeslots_to_conferences.rb
@@ -1,7 +1,7 @@
 class AddAllowedEventTimeslotsToConferences < ActiveRecord::Migration[5.2]
   def up
-    add_column :conferences, :allowed_event_timeslots_csv, :string
-    
+    add_column :conferences, :allowed_event_timeslots_csv, :longtext
+
     Conference.all.each do |conference|
       conference.update_attributes(allowed_event_timeslots: (1..conference.max_timeslots))
     end
--- a/db/migrate/20200915222257_allowed_event_timeslots_length.rb
+++ b/db/migrate/20200915222257_allowed_event_timeslots_length.rb
@@ -1,0 +1,5 @@
+class AllowedEventTimeslotsLength < ActiveRecord::Migration[5.2]
+  def change
+    change_column :conferences, :allowed_event_timeslots_csv, :longtext
+  end
+end


eladeyal-intel:
This edits #651 and nominally should be a new migration. I'm not using/testing frab w/ mysql though. If this helps I can reformat this PR as migration.

manno:
> This edits #651


The migration from #651 will not work on mysql with existing data.

The data migration will raise an exception and leave the deployment in a broken state:
https://github.com/eladeyal-intel/frab/blob/144c777159517149598808fe3bfa2b36832bb78c/db/migrate/20191116151919_add_allowed_event_timeslots_to_conferences.rb#L6


In theory disabling strict mode (?) for the DB could help, but would create truncated CSVs.
I can't think of any other way to fix that, than to edit the original migration.

eladeyal-intel:
Ahh I see. The problem is that the migration will fail in case an existing conference had too many [potential] options for populating `allowed_event_timeslots_csv`. I guess this workaround will do fine. I think you should also edit https://github.com/frab/frab/blob/master/db/schema.rb (by running.... `rails db:schema:dump` if I correctly recall )

eladeyal-intel:
I think `:longtext` is DB-specific and not ActiveRecord type. MyuQL `LONGTEXT` in 4GB long (https://guides.rubyonrails.org/v3.2/migrations.html#supported-types). MySQL `TEXT` is 64KB which should be enough for any conference.

I think `:text` will map to `TEXT` - https://github.com/rails/rails/blob/46ea76980f8d335bc723dedcd6272eac0e8582d3/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L104

eladeyal-intel:
I think `:string` is limited to 255 characters, which gives approx 80 options (i.e., `1,2,3,4,.....,85` is 243 characters). Aren't 80 options enough for every sane conference? i.e, if the slot duration is  1 minute it will let one specify events ranging from 1min to 1:20minutes; and if the slot duration is 5 minutes (which is more plausible) it will let one specify events ranging from 5min to 7h. This field only defines values which are available for submitter to specify which event length they're after. It does not affect events already submitted (or presented). So supporting more than 80 options is not such a good solution anyway.

I propose to change the migration code to add a special case, if `conference.max_timeslots` > 70, then the allowed allowed_event_timeslots_csv will be a sane default, i.e. = `conference.default_timeslots`. So the migration will be changing the UI only for conferences which are open for submissions and have more than 70 timeslots. This way we can keep the `:string` which I imagine is more performant.

eladeyal-intel:
I thought about it some more and decided :text would be good enough. If we want to keep :string we'll need to add validations whenever editing the conference settings too.

eladeyal-intel:
See #726

eladeyal-intel:
In #726 I solved this issue, including fixing a failed mysql migration which may have left the DB in an non consistent state

eladeyal-intel:
I merged #726 so this can be closed

PR #726, opened 2020-09-18
Longer allowed_event_timeslots_csv db field (v2)

eladeyal-intel:


eladeyal-intel:
The maximum for this field is 1,2,...,100 which is 291 characters. Replacing `:string` to `:text` means max number of characters (in MySQL) will be 4K instead of 255.

This is my take on #724 , @manno ; Unfortunately I cannot test it locally on MySQL; it does not work for me with WSL.

eladeyal-intel:
Thanks to https://www.db4free.net I could validate this on mysql too

saerdnaer (approved):
LGTM

PR #748, opened 2020-11-07
Preserve deleted attachments

eladeyal-intel:
If the environment variable `FRAB_PRESERVE_FILE_ATTACHMENTS` is set to 1, frab will not delete a file attachment when it is removed from an event. The former attachment will still be accessible to admins via the history views. To delete all such files from disk (to reclaim disk space and respect submitters' expectation of privacy), admin can use `rake frab:delete_removed_files` which will remove all files which were deleted more than a specified number of days ago.


erdgeist:
This better be communicated to the submitter that their document still remains on the server even after they deliberately asked for it to be removed. I'd even think that the submitter should have a right to override the server's setting.

What is your specific use case to want it to remain?

eladeyal-intel:
We use the event space as a common space between submittors and conference crew. By preserving old attachments, crew is able to have visibility to history of submission.

erdgeist:
But why would someone want to delete an attachment if it was not actually meant to be deleted? And in that case: Why would your org override that decision?

eladeyal-intel:
He submitted the slides for the events. Feedback provided. He should now upload the new version and remove (maybe: archive) the old one.

erdgeist:
Okay, I can see that. Still in that case, the submitter should be informed that the original attachment will not actually be deleted so they can take additional steps if they contain sensitive material uploaded by accident.

eladeyal-intel:
image

erdgeist (approved):

eladeyal-intel:
Thank you @erdgeist

eladeyal-intel:
@erdgeist @manno @saerdnaer  - I think this PR is now ready. I will merge it in a week or so if no feedback is provided.

eladeyal-intel:
https://github.com/frab/frab/wiki/Manual/_compare/87da22033c18dae92c3ac564d859f5f0c4ff727d...a1203ed1a402c44a90391b225a3f4be338b1ea2f

PR #756, opened 2020-12-05
Hotfixes for prod

manno:


eladeyal-intel:
Env cars must be lowercase true in order to be admitted? Also some documentation would be nice.

manno:
I added some docs to the env.example file. I'm just doing this to get rid of a (too) complicated deployment branch.

PR #764, opened 2021-01-01
add ruby 3.0 to CI

eladeyal-intel:


saerdnaer (approved):

eladeyal-intel:
I think ruby 3 is not supported by rails 5

Issue #772, opened 2021-01-24
hide description field


eladeyal-intel:
Some conference owner would prefer shorter forms and rely on submisission data in attached file. So I'd like to enable per-conference configuration option to remove the "long description" field from CFP and Event editing forms.

PR #787, opened 2021-02-20
add DISABLE_FRAB_DESCRIPTION_FIELD

eladeyal-intel:


eladeyal-intel:
Allows hiding the description field so that CfP is shorter. This is related to #772 but it's per-deployment instead of per-conference.

saerdnaer (approved):
LGTM

PR #785, opened 2021-02-20
remove EOL action

eladeyal-intel:

PR #788, opened 2021-02-25
remove action which is EOL

eladeyal-intel:


eladeyal-intel:
`actions/setup-ruby@v1` provides the following warning:

```
NOTE: This action is deprecated and is no longer maintained.
Please, migrate to https://github.com/ruby/setup-ruby, which is being actively maintained.
```

See also https://github.com/actions/setup-ruby


The new action, `ruby/setup-ruby@v1` also includes the gemfil isntallation with caching.

Time is also reduced from 17s to 7 seconds or so.

(same as #785 , something is wierd with CI)

PR #801, opened 2021-04-02
bump rails and mimemagic

eladeyal-intel:
`mimemagic` version 0.3.5 was yanked, and `rails` was upgraded to stop using `mimemagic` altogether. We still need `mimemagic` for `kt-paperclip` though (See https://github.com/kreeti/kt-paperclip/pull/54) but at least it builds now.


eladeyal-intel:
I think this may cause some files to not be offline identified when uploaded to frab. This will be eventually fixed by marcel.

saerdnaer (approved):
LGTM

PR #808, opened 2021-04-24
Numeric filter appearance

eladeyal-intel:
For English, change the labels to words instead of math symbols

image

For all other locales, keep the old appearance until someone contirbutes a translation

image

Also add a test


saerdnaer (approved):
LGTM

PR #870, opened 2021-10-20
Upgrade to Rails 6.1 and Ruby 3.0.2

manno:
I don't have time to pursue this at the moment.

- [x] Prawn
- [x] Tests and CSS matchers
- [x] Static program export (templates)
- [x] ...?


eladeyal-intel:
@manno  the sqlite CI failure can be removed by changing `=9` to `=4` in `.github/workflows/ci.yaml`  line 83.

The idea of that part of the test is - we check we can rollback a few DB migrations, then migrate, and we want to verify that the identical schema was reached (`db/schema.rb` is created with sqlite)

It seems something changed - the offending migrations did not change, and they had `foreign_key: true` but now the add this to the schema file

```
add_foreign_key "average_review_scores", "events"
add_foreign_key "average_review_scores", "review_metrics"
add_foreign_key "review_metrics", "conferences"
add_foreign_key "review_scores", "event_ratings"
add_foreign_key "review_scores", "review_metrics"
```

So I'm not 100% if this reveals a real problem; but anyway it's a specific key to check migrations, and those surely did not become any worse.

.github/workflows/ci.yaml
```diff
@@ -78,12 +78,7 @@ jobs:
     - name: Run tests
       run: |
         sed -i 's/config.eager_load = .*/config.eager_load = true/' config/environments/test.rb
-        bundle exec rails db:schema:load --trace
-        mv db/schema.rb db/schema.rb.original
-        bundle exec rails db:rollback STEP=4
-        bundle exec rails db:migrate
-        sh -c "[ '$DB' != 'sqlite' ] || diff db/schema.rb db/schema.rb.original "
-        bundle exec rails db:test:prepare
+        bundle exec rails db:setup
```
eladeyal-intel, line 82: @manno   

Comment regarding https://github.com/frab/frab/pull/870/commits/7fc92db9857a1aea36bac8872e8b376950060cbd


I think we should just remove the `diff` lines. The rollback+migrate test can stay. They are useful for making sure the migrations actually work.

eladeyal-intel: it seems to be fixed now. thanks

PR #878, opened 2021-12-27
update migration & rollback test

eladeyal-intel:
Current implementation no longer works for Rails 6. The new implementation now
would work on all databases in CI.


eladeyal-intel:
ohhh nevermind. closing

PR #1029, opened 2026-05-01
Improve error handling and feedback spam prevention

manno:
Handle BadRequest and ParameterMissing errors with clearer logging and responses. Add invisible captcha to feedback form and controller to reduce spam. Optimize event conflict updates to avoid unnecessary processing. Refactor exception notification rate limiting in production.


manno (commented):
👍

